PR #101, opened 2026-05-15
fix(ci): contenttypes_live PG test missing _pool→Pool migration from #98

ujeenet:
PR #98 renamed every ContentType `*_pool` method to drop the suffix and removed the PG-only `&PgPool`-taking variants. The `#[cfg(feature = \"postgres\")]`-gated `tests/contenttypes_live.rs` wasn't included in the rename sweep — it was still passing `&PgPool` to functions that now expect `&Pool`.

Result: every PR since #98 merged has had `postgres_test` failing on this file, blocking CI green even when the PR's own code is clean.

## Fix

Every ContentType / GenericForeignKey call site that takes the pool now uses `&pool.clone().into()` (`From<PgPool> for Pool` is in scope). 23 call sites updated. No semantic change — the tests still run against Postgres, the conversion is free.

## Verification

`cargo build --test contenttypes_live --features tenancy,postgres` clean locally.

## Why a separate PR

This is a latent regression from #98 that's affecting every open PR's `postgres_test`. Shipping it as its own small PR means it can merge fast and unblock all the others without each downstream PR carrying the fix.

### DIFF
--- a/crates/rustango/tests/contenttypes_live.rs
+++ b/crates/rustango/tests/contenttypes_live.rs
@@ -187,7 +187,7 @@ async fn ensure_seeded_inserts_a_row_per_model() {
         eprintln!("DATABASE_URL unset — skipping");
         return;
     };
-    let inserted = contenttypes::ensure_seeded(&pool)
+    let inserted = contenttypes::ensure_seeded(&pool.clone().into())
         .await
         .expect("ensure_seeded");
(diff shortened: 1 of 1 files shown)


ujeenet:
Superseded by #102 which includes this fix plus the aggregate decoder fix for PG arrays / jsonb.

PR #104, opened 2026-05-15
fix(test): serialize contenttypes_batch_cache tests on the global cache

ujeenet:
`get_by_natural_key_serves_from_cache_on_repeat` was flaking on CI:

```
thread 'get_by_natural_key_serves_from_cache_on_repeat' panicked:
second call hits cache, no DB:
Driver(Database(SqliteError { code: 1,
  message: \"no such table: rustango_content_types\" }))
```

## Root cause

The ContentType cache is **process-global** (`static OnceLock<RwLock<HashMap<...>>>`). Cargo runs tests in parallel by default. The failing test:

1. `clear_cache()` + populate
2. drop the table
3. assert HIT (expects cache to serve)

Between steps 1 and 3, another test in the same binary calls `clear_cache()`. The entry evicts, step 3 hits the DB, the table is gone — driver error.

It passed locally and on previous CI runs only because the schedule happened to interleave favourably.

## Fix

Suite-wide `tokio::sync::Mutex<()>` (`cache_lock()`); every test takes the lock at entry. Guard lives until end of scope — Rust drops it automatically. Tokio's `Mutex` is required (not `std`'s) because the test crosses `.await` points.

```rust
fn cache_lock() -> &'static Mutex<()> {
    static M: OnceLock<Mutex<()>> = OnceLock::new();
    M.get_or_init(|| Mutex::new(()))
}

#[tokio::test]
async fn whatever() {
    let _g = cache_lock().lock().await;   // ← first line of every test
    contenttypes::clear_cache();
    // ...
}
```

## Verification

5/5 consecutive default-parallelism runs pass:
```
$ for i in 1..5; cargo test --features tenancy,sqlite --test contenttypes_batch_cache
test result: ok. 8 passed; 0 failed
test result: ok. 8 passed; 0 failed
test result: ok. 8 passed; 0 failed
test result: ok. 8 passed; 0 failed
test result: ok. 8 passed; 0 failed
```

## Memory note

Captured the pattern in `feedback_test_global_state.md` — every test mutating process-global state (caches, signal registries, env vars, anything in `static OnceLock<Mutex<...>>`) needs the lock proactively. Adjusted `MEMORY.md` to surface it.

### DIFF
--- a/crates/rustango/tests/contenttypes_batch_cache.rs
+++ b/crates/rustango/tests/contenttypes_batch_cache.rs
@@ -1,12 +1,33 @@
 //! Tests for `ContentType::get_for_models` batch lookup and the
 //! `get_for_model` / `get_by_natural_key` cache layer
 //! (issue #35). Runs against in-memory SQLite — no infra needed.
+//!
+//! ## Why the suite-wide serializing mutex
+//!
+//! The ContentType cache (`contenttypes::clear_cache` / the static
+//! HashMap behind `get_*` methods) is **process-global**. Under
+//! cargo's default parallel test harness, two tests racing on
+//! `clear_cache()` between another test's "populate" and "assert HIT"
+//! calls would evict the entry and force a DB hit — which then fails
+//! on a dropped table in the table-drop tests below. The lock makes
+//! every test in this file run sequentially against the shared cache.
 
 #![cfg(feature = "sqlite")]
+
+use std::sync::OnceLock;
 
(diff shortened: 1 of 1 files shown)


ujeenet:
Superseded by #102. The cache test serialization commit (4917d0d) was cherry-picked into #102 so all three CI fixes can land together — none of these PRs could pass CI standalone because each depended on the others.

PR #556, opened 2026-06-01
fix(tenancy): tri-dialect bootstrap + MySQL TEXT/BLOB/JSON defaults

### LINKED ISSUE #1: [orm] F() expressions for column-vs-column comparisons + atomic updates
Django ref: https://docs.djangoproject.com/en/6.0/ref/models/expressions/#f-expressions

## What's missing
`F()` lets the ORM refer to a model field by name inside a query expression. Enables atomic counter updates (`update(views=F('views') + 1)`), column-vs-column compares (`filter(price__gt=F('cost'))`), arithmetic in `update()`, bitwise ops, and date arithmetic via timedelta.

## Current rustango state
No `F` type or column-reference expression in `crate::core::query`. `UpdateBuilder::set_typed` exists but only accepts a literal value.

## Why it matters
Foundational for concurrency-safe write paths. Without it, every counter increment is a read-modify-write race. Also blocks F-derived CheckConstraints.

## Acceptance
- `F("col")` expression that lowers to a quoted-identifier column ref.
- Extend filter rhs + `set_typed` to accept `F` and arithmetic over `F`.
- Bitwise ops (`bitand`/`bitor`/`bitxor`/`bitleftshift`/`bitrightshift`).
- Tri-dialect emit; concurrent-update race-resistance integration test.

Part of the **ORM Expression DSL** epic.

ujeenet:
Two database-mode tenancy bugs, surfaced while bringing **rustango-cms** up on MySQL 8 (the tenancy stack had only been live-tested on Postgres schema-mode). Both also affect *any* 0.42 app using roles/permissions in database-mode tenancy.

## 1. Bootstrap declares roles tables but never creates them
`tenant_bootstrap_migration_for` builds `full_snapshot_for` (declaring `rustango_roles / _role_permissions / _user_roles / _user_permissions / _api_keys`), but its `forward` ops only `CreateTable("rustango_users")`. So the five tables are **declared-as-existing in the snapshot yet never created**. Any downstream migration that trusts the snapshot and FKs to them then:
- **MySQL/db-mode:** fails at apply (`1824: Failed to open the referenced table 'rustango_roles'`)
- **SQLite/db-mode:** silently dangles (FK enforcement off)
- **Postgres schema-mode:** masked (shared DB happens to have them)

Fix: emit the missing `CreateTable` forward ops. The runner defers FKs, so creation order is irrelevant.

## 2. MySQL literal DEFAULT on TEXT/BLOB/JSON (err 1101)
MySQL rejects `DEFAULT '{}'` on TEXT/BLOB/JSON columns. `translate_default_expr` only parenthesized `json`; extend to `string` (→ TEXT) and `binary` (→ BLOB) via the 8.0.13+ expression-default form `DEFAULT (<expr>)` (a harmless wrap for bounded `VARCHAR`).

## Verification
A rustango-cms tenant (users + roles + 42 cms tables) now migrates → seeds → serves (admin login + sitemap querying content) on **both SQLite and MySQL 8** database-mode. SQLite was already green; these two fixes close MySQL.

⚠️ CI should run the live `*_mysql_live` / `*_sqlite_live` + `migrate::ddl` suites — fix #1 changes what the tenant bootstrap creates, so any test that provisioned roles by other means may need a touch-up.

### DIFF
--- a/crates/rustango/src/sql/mysql.rs
+++ b/crates/rustango/src/sql/mysql.rs
@@ -184,10 +184,12 @@ impl Dialect for MySql {
     ///   to carry `(6)` too, otherwise MySQL fails with
     ///   `1067 (42000): Invalid default value`.
     /// - `'<lit>'::<type>` → `'<lit>'` (strip Postgres cast).
-    /// - JSON columns (`ty == "json"`): wrap the result in parens —
-    ///   `DEFAULT '{}'` is rejected by MySQL on JSON/TEXT/BLOB
-    ///   (`1101 (42000)`), but the MySQL-8.0.13+ expression-default
-    ///   form `DEFAULT (<expr>)` is accepted.
+    /// - TEXT / BLOB / JSON columns (`ty` ∈ {`json`, `string`,
+    ///   `binary`}): wrap the result in parens — `DEFAULT '{}'` is
+    ///   rejected by MySQL on JSON/TEXT/BLOB (`1101 (42000)`), but the
+    ///   MySQL-8.0.13+ expression-default form `DEFAULT (<expr>)` is
+    ///   accepted (and is a harmless wrap for bounded `string` →
+    ///   VARCHAR).
     /// - Everything else passes through.
     fn translate_default_expr(&self, expr: &str, ty: &str) -> String {
         let mut out = expr.trim().to_owned();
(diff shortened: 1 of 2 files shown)


ujeenet:
Closing as **superseded by develop** — #315 was resolved a different way, ~12 days before this PR was authored.

**Half 1 (bootstrap `CreateTable` ops):** develop creates the roles/permission tables via `tenancy::permissions::ensure_tables_pool` — raw tri-dialect `CREATE TABLE IF NOT EXISTS` DDL wired into the provisioning flow (migrate.rs:507/573/608, landed 2026-05-20 in #315). The merge-base of this branch already has all 6 call sites. Adding `CreateTable` ops to the bootstrap on top of that duplicates the DDL, diverges from `ensure_tables_pool`'s actual table set, and breaks the sibling test `tenant_migration_is_tenant_scoped` (the `forward.len() == 1` → 6 assertion, which is the postgres_test failure here). Folding table creation into the bootstrap migration (and retiring `ensure_tables_pool`) is acknowledged future work in that helper's own doc — but it's a deliberate refactor, not this piecemeal change.

**Half 2 (MySQL TEXT/BLOB/JSON `DEFAULT`):** already fixed on develop, and more precisely. `translate_default_expr` now takes `max_length`, wraps `json`/`binary`/unbounded-`string`(TEXT), and correctly leaves bounded VARCHAR defaults unwrapped. Develop also ships the equivalent unit tests. This PR edits the old 2-arg signature, which is the source of the conflict and would no longer compile against develop's callers.

Develop CI is green on `mysql_live` + `postgres_test`. No regression from closing this. Remaining follow-up is purely verification: re-run the rustango-cms MySQL #315 check against current develop.

PR #608, opened 2026-06-04
fix(test): debounce ttl_purges_lazily_on_read second-boundary race

ujeenet:
## Summary

Surfaced on PR #605's CI run for unrelated index_include changes.

The test set the TTL to **1 second** and immediately asserted the value was present, then slept 1100ms and asserted absence. SQLite's TIMESTAMP column stores **integer-second precision**, so under CI load a `set` at `HH:MM:SS.999` followed by a slow `get` at `HH:MM:SS+1.001` read the expiry as already past — `get` returned `None` where the test expected `Some("v")`, failing the immediate-read assertion.

Bumping the TTL to **3 seconds** (with a 3.2s sleep before the absence assertion) gives ~3000ms of slack before the next second-boundary tick matters. Test semantics unchanged.

## Test plan

- [x] `cargo test --features sqlite,tenancy,cache,config --test cache_db_backend_sqlite_live ttl_purges_lazily_on_read` — green

### DIFF
--- a/crates/rustango/tests/cache_db_backend_sqlite_live.rs
+++ b/crates/rustango/tests/cache_db_backend_sqlite_live.rs
@@ -52,14 +52,20 @@ async fn ttl_purges_lazily_on_read() {
     let cache = DatabaseCache::new(pool, "rustango_cache_ttl");
     cache.ensure_table().await.unwrap();
 
+    // Use a 3-second TTL with a 3.2-second sleep below. Earlier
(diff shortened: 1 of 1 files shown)


ujeenet:
Superseded by the millisecond-precision fix in #609 — the actual race is in the cache backend's integer-seconds storage, not the test's TTL value. Closing in favor of the precision fix that preserves the 1-second TTL test semantics.

PR #1125, opened 2026-07-10
fix(examples): revive cookbook_blog + CI-gate it + QUERY chapter (#1124, #1114)

### LINKED ISSUE #1114: [query] Slice 7 — OpenAPI 3.2 query operation + docs + cookbook recipe
Part of #1107 (HTTP QUERY epic). Final slice — depends on all others.

## OpenAPI

- Our emitter pins `openapi: "3.1.0"` (`openapi/mod.rs:68`); OpenAPI 3.1 has a fixed operation set with no QUERY. [OpenAPI 3.2](https://www.openapis.org/blog/2025/09/23/announcing-openapi-v3-2) (Sept 2025) added a first-class `query` operation on the Path Item plus `additionalOperations`.
- Plan: support `query` operations in the schema builder and bump the emitted version to `3.2.0` **when a spec contains QUERY ops** (specs without them can stay 3.1.0 for maximum tooling compat — decide in-issue whether conditional or unconditional bump is less surprising).
- Verify the Swagger-UI/redoc versions referenced in our docs render 3.2 `query`; if the ecosystem isn't ready, document the limitation rather than blocking the emit.
- QUERY request bodies documented via standard `requestBody` on the operation.

## Docs

- New `docs/query-method.md`: what QUERY is (RFC 10008), when to reach for it vs GET (URL-length limits, complex/structured criteria, cacheable searches), curl examples (`curl -X QUERY --data …`), and the framework guarantees: CSRF exemption rationale, CORS preflight behavior, cache_page opt-in, retry semantics.
- README: docs-section link + What's new entry.
- Cookbook: recipe chapter — product-search endpoint accepting structured JSON criteria via QUERY with the identical handler also serving `GET ?q=…`, tested through `TestClient::query`.

## Acceptance

- `paths./search.query` appears in generated spec with requestBody schema.
- Docs build clean; cookbook example compiles and its tests pass.

### LINKED ISSUE #1124: [examples] cookbook_blog is bit-rotted + CI-untested — revive it and add to doc_examples CI
The `crates/rustango/examples/cookbook_blog` example (the 16-chapter `COOKBOOK.md`, ~3,240 lines) **does not compile** and is **not covered by CI**, so it has silently bit-rotted against several rustango API changes. This blocks #1012 (add missing chapters) and #1114 (QUERY recipe chapter) — both add cookbook chapters, which can't be verified while the example is broken.

## Why CI is green despite this

The `doc_examples` matrix in `.github/workflows/ci.yml` tests `orm_cookbook`, `getting_started_blog`, `admin_demo`, `auth_demo` — **not** `cookbook_blog`. So nothing gates it and the rot went unnoticed.

## Root cause #1 (blocker) — missing `[features]` section

`cookbook_blog/Cargo.toml` has **no `[features]` section**. `#[derive(Model)]` emits backend-specific `sqlx::FromRow` impls gated on the *deriving* crate's own features (`#[cfg(feature = "postgres")]`, see `crates/rustango-macros/src/lib.rs:1354`). With no `postgres` feature on the example crate, those impls compile out and every model fails `MaybePgFromRow` / `FromRow<PgRow>` (e.g. `QuerySet<Author>::fetch` "trait bounds not satisfied").

**Fix** (mirrors the working `orm_cookbook`):
```toml
[features]
default = ["postgres"]
postgres = ["rustango/postgres"]
mysql = ["rustango/mysql"]
sqlite = ["rustango/sqlite"]
```

## Root cause #2 — API bit-rot (revealed once #1 is fixed)

~36 errors across 7 chapter test files, in 3 mechanical categories:

1. **Old `filter` signature** (10 sites) — `.filter("id", Op::Eq, id)` (3-arg) → `.filter("id", id)` (Eq implied) or `.filter_op("id", Op::Eq, id)`. Sites: `src/apps/blog/urls.rs:40,139`, `tests/cookbook_chapter02_models.rs`, `chapter03_orm.rs`, `chapter12_bidialect.rs`.
2. **`PgPool` → `rustango::sql::Pool`** (16 sites, v0.36 multi-backend migration) — rustango APIs now take the backend-erasing `sql::Pool` enum; chapters pass a raw `sqlx::PgPool` (`expected Pool, found Pool<Postgres>`). Fiddly per-chapter: each `fn pool() -> Option<sqlx::PgPool>` is used BOTH for raw `sqlx::query(...)` setup AND for rustango APIs, so the two usages must be threaded separately (wrap with `Pool::from(..)` at the rustango call sites). Chapters: `05_tenancy`, `07e_unique_together_validator`, `09c_template_views`, `06b_user_extractors`.
3. **`TenantContext` field rename** (1 site) — `chapter09d_viewset_tenant_router.rs:151` uses `TenantContext { registry: ... }`; the `registry` field no longer exists.

## Recommended fix (one PR)

1. Add the `[features]` section (root cause #1).
2. Fix the 3 API-drift categories above (compiler-verified).
3. **Add `cookbook_blog` to the `doc_examples` CI matrix** (`migrate: false`, Postgres service already provided) so it can never silently rot again. This is the load-bearing part — without it the example will re-rot.

Verification needs Postgres (the chapters are `_live` tests that skip when `DATABASE_URL` is unset); local `cargo test --no-run` verifies compilation, CI verifies runtime.

Once this lands, #1114's QUERY chapter (drafted, `chapter09e_query_search`) and #1012's missing chapters can be added and verified.

ujeenet:
Closes #1124. Closes #1114.

The `cookbook_blog` example (the 16-chapter `COOKBOOK.md`) had silently bit-rotted — it didn't compile and **wasn't in CI**. This revives it, gates it in CI so it can't rot again, and adds the QUERY recipe chapter (#1114) that was blocked on it.

## Root cause + fix

`cookbook_blog/Cargo.toml` had **no `[features]` section**, so `#[derive(Model)]`'s `#[cfg(feature = "postgres")]`-gated `FromRow` impls compiled out and every model failed `MaybePgFromRow`. Added the section (`default = ["postgres"]` + forwards), mirroring `orm_cookbook`.

Then the usual API bit-rot (all compiler-verified):
- **filter API** — `.filter(k, Op::X, v)` (3-arg) → `.filter_op(k, Op::X, v)` (19 sites, ch02/03/12).
- **`PgPool` → `sql::Pool`** (v0.36 multi-backend) — `&pool.clone().into()` at rustango-API call sites, keeping raw `sqlx` uses on `&PgPool` (ch02/05/06b/07e/09c).
- `fetch_aggregate`→`fetch_aggregate_on`, `bulk_insert`→`bulk_insert_on` (ch03).
- dropped the removed `TenantContext.registry` field (ch09d — the registry pool lives in `pools` now).
- fixed a stale embedded-migration-name assertion (ch04: `0002_cookbook_tables`, not `0001_initial`).
- removed 3 now-unused imports.

## Durable fix

**Added `cookbook_blog` to the `doc_examples` CI matrix** (`migrate: false`; Postgres service already provided) — this is the part that prevents recurrence.

## New — Chapter 9e (closes #1114)

The RFC 10008 QUERY search recipe: one `get(search).query(search)` handler over `Params<ProductQuery>` serving `GET ?q=…` and `QUERY` (urlencoded + JSON arrays), driven by `TestClient::query`. **No DB** — runs everywhere. `COOKBOOK.md` updated with the 9e section.

## Verification

Compiles clean (0 warnings); all non-DB chapters pass locally (incl. the new 9e). The `_live` chapters need Postgres, so **CI is the runtime gate** — the new matrix entry runs them against the CI Postgres service. Merging only after CI is green.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -486,8 +486,13 @@ jobs:
           # The 16-chapter COOKBOOK.md example. Chapters manage their own
           # DB setup (reset/migrate per test), so migrate: false like
           # orm_cookbook. Gated here so it can't silently bit-rot (#1124).
+          # `--test-threads=1`: the chapters create/reset tables in the
+          # shared `public` schema, so they must run serially (as every
+          # chapter's doc comment already documents) — parallel DDL races
+          # on `pg_class`/`pg_type`.
           - example: cookbook_blog
             migrate: false
+            test_args: "-- --test-threads=1"
           - example: admin_demo
             migrate: true
           # Authentication deep-dive docs (docs/auth-*.md). `migrate: false` for
@@ -526,4 +531,4 @@ jobs:
         run: cargo run -- migrate
       - name: Test
         working-directory: crates/rustango/examples/${{ matrix.example }}
-        run: cargo test
+        run: cargo test ${{ matrix.test_args }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -483,6 +483,11 @@ jobs:
             migrate: true
           - example: orm_cookbook
             migrate: false
+          # The 16-chapter COOKBOOK.md example. Chapters manage their own
+          # DB setup (reset/migrate per test), so migrate: false like
+          # orm_cookbook. Gated here so it can't silently bit-rot (#1124).
+          - example: cookbook_blog
+            migrate: false
           - example: admin_demo
             migrate: true
           # Authentication deep-dive docs (docs/auth-*.md). `migrate: false` for
--- a/crates/rustango/examples/cookbook_blog/COOKBOOK.md
+++ b/crates/rustango/examples/cookbook_blog/COOKBOOK.md
@@ -2339,6 +2339,42 @@ gone.
 9.116b (typed permissions), 9.117 (OpenAPI auto-derive),
 9.118 (response shaping via `.fields(&[...])`) queued for Slice 9b.*
 
+## Chapter 9e — Searching with the HTTP `QUERY` method (RFC 10008)
+
+3 tests, **no DB** — one product-search handler serving both `GET` and
+`QUERY` from the same code. Run with
+`cargo test --test cookbook_chapter09e_query_search`.
+
+`QUERY` is the "safe GET with a body": safe + idempotent like `GET`, but
+the search criteria travel in the request body — so a search too big or
+too structured for a querystring (long filter lists, arrays, nested
+criteria) doesn't have to masquerade as a `POST`. Two pieces combine:
+`QueryRouterExt::query` routes `QUERY` alongside `GET` on one path, and
+the `Params<T>` extractor reads `T` from the querystring on `GET` and
+from the body on `QUERY`.
+
+```rust,ignore
+use rustango::params::Params;
+use rustango::http_query::QueryRouterExt;
+use axum::routing::get;
+
+async fn search(Params(q): Params<ProductQuery>) -> Json<Value> { /* filter */ }
+
+Router::new().route("/products", get(search).query(search));
+```
+
+* §9.119 — `GET /products?q=cable&max_price=15` and the same criteria in
+  a urlencoded `QUERY` body return identical results.
+  → `get_and_query_urlencoded_agree`
+* §9.119 — a `QUERY` JSON body carries a real array (`{"tags":["usb","hdmi"]}`)
+  — awkward in a querystring, natural in a body. → `query_json_body_matches_multiple_tags`
+* §9.119 — an empty `QUERY` body returns the whole catalog.
+  → `empty_query_returns_everything`
+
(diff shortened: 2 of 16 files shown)


ujeenet:
**Scope note:** verifying the revived cookbook against a real Postgres (via a local Docker PG matching the CI env) surfaced runtime issues a local compile couldn't, so this PR now also includes a small **framework fix**: `validate_unique_together`'s pre-check ran `SELECT 1 … LIMIT 1` decoded as `(i64,)`, which fails on PG (the literal `1` is INT4; the only test was SQLite-only). Fixed to `SELECT COUNT(*)` (bigint everywhere). Plus ch03 scalar-aggregate `.values(&[])`, the v0.29 `/__login`→`/login` & `/__admin`→`/admin` route-flip in the booted-binary chapters, an 08b DROP-DATABASE connection-termination fix, and `--test-threads=1` for the cookbook CI job. **Full suite now passes against Postgres: 120 tests, 0 failures** (verified locally).

Issue #1161, opened 2026-07-19
migrate: `#[rustango(default = "")]` renders invalid DDL (empty-string default → `near "NOT": syntax error`)


ujeenet:
## Summary
A model field with an **empty-string default** (`#[rustango(default = "")]`) generates malformed `CREATE TABLE` DDL: the empty string renders as *nothing* after the `DEFAULT` keyword, so the column reads `… DEFAULT  NOT NULL` and the driver rejects it with `near "NOT": syntax error`.

## Repro
Add a string-typed field with an empty-string default to any model, e.g.:

```rust
#[derive(rustango::Model)]
#[rustango(table = "demo")]
struct Demo {
    #[rustango(primary_key)]
    id: rustango::sql::Auto<i64>,
    #[rustango(max_length = 64, default = "")]
    name: String,
}
```

Then generate + apply the migration (sqlite):

```
Error: Migrate(Driver(Database(SqliteError { code: 1, message: "near \"NOT\": syntax error" })))
```

Discovered while adding `client_secret: Cast<EncryptedString>` with `#[rustango(default = "")]` to `SsoProvider` (`admin/sso_provider.rs`) — `CREATE TABLE rustango_sso_providers` failed on sqlite.

## Root cause
The migration **snapshot is correct** — the field serializes as:
```json
{"name":"name","column":"name","ty":"string","nullable":false,"max_length":64,"default":""}
```
So the bug is in **DDL rendering** (`src/migrate/ddl.rs`, the `DEFAULT` clause / `translate_default_expr`): an empty-string default emits `DEFAULT ` with no value instead of `DEFAULT ''`. Result:
```sql
"name" VARCHAR(64) DEFAULT  NOT NULL   -- ← `near "NOT"`
```

## Expected
An empty-string default should render as a quoted empty literal:
```sql
"name" VARCHAR(64) NOT NULL DEFAULT ''
```
(and equivalently for MySQL/Postgres — quote the empty string).

## Impact
Any `String` / `Cast<C>` field declaring `default = ""` produces an un-appliable migration on a fresh `CREATE TABLE` (not just `AddColumn`). Likely all three dialects, since the empty string is dropped before dialect-specific quoting.

## Workaround
Omit the default (`NOT NULL` with no default, provide the value on insert) or use a non-empty default.

PR #1162, opened 2026-07-19
admin-sso: DB-backed multi-provider SSO + encrypted client secrets

ujeenet:
Lands the DB-backed, admin-managed, **multi-provider** SSO, and stores each provider's client secret **encrypted at rest**.

## Multi-provider (earlier commits on this branch)
Replaces the flat `Org.sso_*` columns with per-tenant `SsoProvider` rows + a registry-wide `SharedSsoProvider` set, merged tenant-wins. Each enabled row is a "Sign in with …" button at `/login/sso/{slug}`.

## Encrypted client secret (this PR's addition)
`SsoProvider` + `SharedSsoProvider` now carry `client_secret: Cast<EncryptedString>` (XChaCha20-Poly1305, key from `RUSTANGO_SECRET_KEY`) instead of a `secret_ref` env/vault handle. Each tenant stores **its own** secret in its own row — no per-tenant env var, no plaintext in the DB. Decrypted in-memory only for the IdP token exchange (sent over TLS, as OAuth requires). `admin-sso` now depends on `casts`.

## Notes
- Requires `RUSTANGO_SECRET_KEY` on the app (the AEAD key).
- Worked around #1161 (empty-string `default = ""` → invalid DDL).
- Verified locally end-to-end via cms_demo: add provider in the admin UI → secret stored as ciphertext in the DB → login button renders → `/login/sso/{slug}` resolves the decrypted secret.


ujeenet:
Wrong base — the default branch is `develop`, not `main` (which is a divergent 0.48 lineage). Superseded by the existing → develop PR.

PR #1182, opened 2026-08-08
feat(viewset): principal-aware filter backends that scope every action

ujeenet:
Two gaps that together make a per-user REST API impossible to build safely on `ViewSet`.

## 1. A backend cannot see who is asking

`ViewSetFilter::filter(params, schema)` receives only the query string. The authenticated principal lives in the request extensions, so the single predicate every per-user API needs — *rows belonging to the caller* — could not be expressed at all.

Added `filter_with(&Parts, params, schema)`, **defaulted to `filter`**, so every existing backend (including the plain closure form) compiles and behaves exactly as before.

## 2. Backends ran on `list` only

This is the worse half. Scoping the collection while leaving every row reachable by id reads as safe and isn't: an app that filtered its list was still one `GET /notes/3` away from another user's data.

Backends now narrow `retrieve`, `update` and `destroy` too — DRF's `get_queryset()` contract:

- an excluded row is **404**, not 403, on the item routes (a 403 confirms the id exists);
- an `UPDATE` matching nothing returns 404 instead of reporting success;
- the read-back after an update is scoped, so it cannot return a row the caller may not see.

## Tests

`crates/rustango/tests/viewset_principal_filter_sqlite_live.rs` — list narrows per principal; cross-owner retrieve/update/destroy are 404 **with the row asserted untouched**; the owner's own update still works and reads back; and a backend with no principal in hand fails closed rather than widening to every row.

`cargo test --workspace --all-features`: 494 test binaries, zero failures. `cargo clippy -p rustango --features tenancy --lib --no-deps` clean.

Version 0.52.0 — new public trait method, plus a behaviour change on the item routes for anyone already using `filter_backend`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,30 @@ All notable changes to rustango. The format follows [Keep a Changelog](https://k
 
 _Nothing yet — next development cycle._
 
+## [0.51.1] — 2026-08-06
+
+Headline: **the media upgrade is now automatic** — the collision that 0.51.0's
+upgrade note warned about (existing `ensure_table`-era media tables vs the new
+system migration) is reconciled during provisioning, with no operator action.
+
+### Fixed
+
+- **Auto-reconcile pre-existing framework tables on provision** (#1167) — the
+  system-migration runner now performs a guarded **fake-initial**: a pending
+  migration that is _purely_ `CreateTable` operations whose tables **all
+  already exist** is recorded in the `__rustango_system_migrations__` ledger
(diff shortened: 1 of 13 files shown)


ujeenet:
Superseded by #1187, which targets `develop` (this one was opened against `main` by mistake) and goes further: the branch now also ships `Principal`, `require_bearer` and the reusable `OwnedBy` backend, so an app names its ownership column and is done.

Issue #1186, opened 2026-08-09
migrate_manage tests fail on a fresh database (ledger-existence assumption in delete_ledger_entry)


ujeenet:
## Symptom

Five tests in `crates/rustango/tests/migrate_manage.rs` fail when run against a **brand-new** database, and pass on every subsequent run against that same database:

```
downgrade_subcommand_steps_back_one_by_default
migrate_dry_run_subcommand_prints_sql_no_writes
migrate_subcommand_applies_pending
migrate_to_target_subcommand_routes_correctly
run_with_writer_captures_migrate_output
```

```
thread '…' panicked at crates/rustango/tests/migrate_manage.rs:99:10
```

## Cause

The `delete_ledger_entry` helper assumes the ledger table already exists (`crates/rustango/tests/migrate_manage.rs:93-100`):

```rust
async fn delete_ledger_entry(pool: &rustango::sql::Pool, name: &str) {
    let pg = pool.as_postgres().expect("test pool is postgres");
    sqlx::query("DELETE FROM __rustango_migrations__ WHERE name = $1")
        .bind(name)
        .execute(pg)
        .await
        .unwrap();          // <- panics: relation "__rustango_migrations__" does not exist
}
```

It runs during test setup, *before* anything has created the ledger. On a database where some earlier test (or an earlier run) already triggered `ensure_ledger`, the table is there and the helper succeeds — which is why it only bites on a fresh DB, and why CI normally stays green (other tests in the parallel workspace run create the ledger first).

## Not a regression

Confirmed pre-existing, not caused by any recent migration work: checking out `origin/develop` at `7b776a8` (v0.51.1) into a clean worktree and running

```
cargo test --all-features --test migrate_manage -- --test-threads=1
```

against a freshly-created Postgres database reproduces exactly the same 5 failures.

## Why it's worth fixing

- Anyone running the suite locally against a new database sees 5 red tests that have nothing to do with their change — it costs time to attribute (it did here, twice, during the #1167 work).
- The suite's green/red result depends on test ordering and on residue in the target database, so it can mask a genuine failure or manufacture a false one.

## Suggested fix

Make the helper self-sufficient rather than order-dependent — either ensure the ledger exists first (`migrate::ensure_ledger_pool`) or treat a missing table as a no-op, since "delete this row if present" is the intent:

```rust
let _ = sqlx::query("DELETE FROM __rustango_migrations__ WHERE name = $1")
[...]

PR #1204, opened 2026-08-13
fix(viewset): page ceiling is the app's, not a hard-coded 1000 (#1196)

### LINKED ISSUE #1196: ViewSet has no per-instance max page_size — the ceiling is a hard-coded 1000
## What happens

`ViewSet::page_size(n)` sets the **default** page size. The runtime clamp is `.min(1000).max(1)` (`viewset/mod.rs:1571`), so any client may ask for up to 1000 rows regardless of what the app configured.

There is no `max_page_size` on the builder. `template_views.rs` does have one, defaulting to 100 — so the two pagination APIs in the same framework disagree by a factor of ten and differ in whether the ceiling is configurable at all.

## Why it matters

Performance, mostly. An app that sets `.page_size(20)` has sized its serializer, its joins and its response budget around 20 rows. A caller passing `?page_size=1000` gets a 50x amplification on all of it — and if the serializer does any per-row work that touches the database, that is where an N+1 turns into a thousand queries in one request.

It is reachable by any authenticated client, and it is the cheapest way to make a rustango app do expensive work.

## Suggested shape

`max_page_size(n)` on the builder, clamping to the app's number rather than to 1000, and defaulting to something closer to `template_views`' 100. Matching the two APIs would be worth doing at the same time.

## Verified in

rustango at the SHA IronBar pins; `crates/rustango/src/viewset/mod.rs:1571` for the clamp, `crates/rustango/src/template_views.rs:292,325` for the contrasting `max_page_size`.

ujeenet:
Closes #1196.

`page_size(n)` set the *default*; the runtime clamp was `.min(1000)`, so any client could ask for 1000 rows regardless of what the app configured. An app sized around `page_size(20)` faced a **50× amplification** of its serializer, joins and response budget — and where the serializer does per-row work against the database, that is an N+1 becoming a thousand queries in a single request. Reachable by any authenticated caller.

### What changed

`max_page_size(n)` on the builder; the clamp uses it instead of 1000.

**Behaviour change:** the ceiling now defaults to **100**, matching `template_views`' existing `max_page_size`. The framework's two pagination surfaces previously disagreed by a factor of ten *and* differed in whether the ceiling was configurable at all — they now agree.

Two bypasses are closed as well:
- `?limit=` is bounded by the same value, so limit/offset isn't a way around it.
- a `page_size` larger than the ceiling is clamped, so it can't smuggle a bigger page through the no-parameter path.

### Tests

The amplification case (`?page_size=1000` → 100), lowering the ceiling, raising it deliberately, `?limit=` honouring it, and the oversized-default path. `limit_offset_clamps_hostile_bounds` encoded the old 1000 contract and was updated to the new one.

Gate: `cargo test --workspace --all-features --no-run` clean, clippy clean, pagination suites green.

### Upgrade note

An app that relied on clients requesting up to 1000 rows should add `.max_page_size(1000)` to restore the old bound.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,18 @@ action, and `OwnedBy` does the common case in one line.
 
 ### Fixed
 
+- **ViewSet page ceiling is the app's, not a hard-coded 1000** (#1196) — a
+  client could ask for `?page_size=1000` regardless of what the app configured,
+  so an app sized around `page_size(20)` faced a 50× amplification of its
+  serializer, joins and response budget; where the serializer does per-row work
+  against the database, that turns an N+1 into a thousand queries in one
+  request. **Behaviour change:** the ceiling now defaults to **100** (matching
+  `template_views`, which the two pagination surfaces previously disagreed
+  about by a factor of ten), and is configurable per ViewSet with
+  `max_page_size(n)`. `?limit=` is bounded by the same value, so limit/offset
+  isn't a way around it, and a `page_size` larger than the ceiling can't
+  smuggle a bigger page through the default path either.
+
 - **ViewSet: 401 for anonymous, 403 for authenticated-but-unauthorised**
   (#1193) — the permission check collapsed both cases to 403. A token client
   treats 401 as its cue to refresh, so answering 403 to a request that simply
--- a/crates/rustango/src/viewset/mod.rs
+++ b/crates/rustango/src/viewset/mod.rs
@@ -39,7 +39,7 @@
 //! | Parameter | Default | Description |
 //! |---|---|---|
 //! | `page` | 1 | 1-based page number |
-//! | `page_size` | configured default | Items per page (capped at 1000) |
+//! | `page_size` | configured default | Items per page (capped at `max_page_size`, default 100) |
 //! | `ordering` | configured default | Comma-separated field names, prefix `-` for DESC |
 //! | `search` | — | Full-text search across `search_fields` |
 //! | `{field}` | — | Exact filter for any `filter_fields` |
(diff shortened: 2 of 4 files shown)


ujeenet:
## Review

The fix is right and the tests are the good kind — they assert row counts from
a live 250-row table rather than inspecting the clamp arithmetic, so they would
survive a refactor of where the clamping happens. Clamping the *default* as well
as the request closes the obvious bypass, and doing `?limit=` with the same
ceiling closes the other one. Three findings, none blocking.

### 1. A third pagination surface still hard-codes 1000 — low

The PR says "the framework's two pagination surfaces now agree". There is a
third: `crates/rustango/src/pagination.rs:187`

```rust
const MAX_PAGE_SIZE: usize = 1000;
```

used by `page_number_links()` (:255) and `cursor_links()` (:283). It is not
configurable at all.

Low severity, and **not a hole in this fix**: those functions only build URL
strings — they fetch no rows, so none of the amplification #1196 is about
applies. But it means a `pagination::page_number_links(…, 1000, …)` caller emits
links advertising a page size the ViewSet will now refuse to serve. Worth a
follow-up for consistency rather than widening this PR.

### 2. Cursor pagination's clamp is claimed but not tested — low

The doc table now says cursor `page_size` is "capped at `max_page_size`, default
100", and the new test file covers page-number, limit/offset, the raised
ceiling, the lowered ceiling and the clamped default — but not cursor. The clamp
is computed once at the top of `run_list` and shared by all three styles, so it
almost certainly holds; the point is that the table makes a claim no test backs,
which is how the doc/behaviour drift in #1186 started.

One more case mirroring `client_cannot_exceed_the_default_ceiling` with
`.cursor_pagination()` would close it.

### 3. `#[must_use]` is inconsistent between the two adjacent builders — nit

`max_page_size` gets `#[must_use]`; `page_size` right above it still doesn't.
Both are `self`-consuming builders where dropping the result is silently wrong.

### Upgrade note worth adding

The ceiling change is **silent** for clients: a caller that hard-codes
`?page_size=500` now receives 100 rows with no error. The response echoes
`page_size`, so a client that reads it paginates correctly, but one that assumed
its request was honoured will under-fetch and may conclude the collection is
smaller than it is. The CHANGELOG calls out the behaviour change; naming that
specific failure mode would help anyone upgrading.

PR #1205, opened 2026-08-13
feat(jwt)!: make JtiStore async so a durable store can just write (#1191)

### LINKED ISSUE #1191: JtiStore is a sync trait, which forces eventual consistency on durable stores
## What happens

The `JtiStore` trait is synchronous, so an implementation cannot await a database write.

## Why it matters

Any durable store has to become a hot `RwLock<HashMap>` plus a background flusher to a registry table — eventually consistent, with a convergence window (~10s in my sketch) during which a revoked jti is still accepted on another instance.

That is a real design constraint imposed by a trait signature rather than by the problem. Revoking a token is exactly the operation you want to be durable and immediate.

Single-instance deployments are unaffected, which is why I have not built it — but it is the thing standing between this app and horizontal scaling.

## Suggested shape

An async trait (or an async variant), so a store can be a straightforward table write.

ujeenet:
Closes #1191.

## The problem the signature created

`JtiStore` was synchronous. A store is consulted on the verify path, which is
async, so a shared multi-instance store could not simply `await` a write — the
only shape that fit a sync trait was a hot in-memory map plus a background
flusher. That is eventually consistent: after `/logout`, the revoked `jti`
stayed **valid on other replicas** for the length of the convergence window.
Revocation is precisely the operation that wants to be durable and immediate,
and the constraint came from the trait signature rather than from the problem.

## What changed

`is_used` / `mark_used` / `approx_size` return `JtiFuture<'_, T>`. Everything
that consults the store is async with them:

- `JwtLifecycle::{verify_token, verify_access, verify_refresh, refresh, refresh_with, revoke, blacklist_jti, is_blacklisted, blacklist_size}`
- `mcp::verify_agent_token`
- `tenancy::auth_routes::verify_for_tenant`
- `tenancy::admin::redeem_impersonation_handoff`

A durable store is now one query:

```rust
fn mark_used<'a>(&'a self, jti: &'a str, exp_unix: i64) -> JtiFuture<'a, bool> {
    Box::pin(async move {
        // INSERT … ON CONFLICT DO NOTHING — atomic, one round trip,
        // visible to every replica immediately.
        self.insert_if_absent(jti, exp_unix).await
    })
}
```

`mark_used` must still be **atomic** — concurrent callers for the same `jti`
must see exactly one `true`, expressed as a single conditional write
(`ON CONFLICT DO NOTHING`, Redis `SET NX`), never read-then-write. The doc
comment says so, and there is now a test that proves the contract holds for a
store that genuinely suspends between the check and the record.

## Why boxed futures and not `#[async_trait]`

Every consumer holds the store as `Arc<dyn JtiStore>`, and native `async fn` in
traits is not dyn-compatible, so the future has to be boxed either way. It is
written out by hand rather than via the macro because `jti_store` is a **core,
ungated** module while `async-trait` is an *optional* dependency, pulled in only
by `cache` / `storage` / `jobs` / `oauth2` / `email` / `secrets`. Using the
macro would break supported feature combinations — verified against the litmus
`--no-default-features --features sqlite,tenancy`, which builds clean.

## Behaviour

- `InMemoryJtiStore` is unchanged in behaviour. Its bodies stay synchronous, so
  the `std::sync::Mutex` guard never crosses a suspend point.
- Token **expiry is now checked before** the store is consulted — an expired
  token costs no round trip.
- Breaking for callers only in that `.await` is required; a synchronous custom
  store stays a one-line `Box::pin(async move { … })` wrapper.

## Verification

- `cargo test --workspace --all-features --no-run` — finished clean
- `cargo build --no-default-features --features sqlite,tenancy` — the combo that
  justified hand-rolling
- `jti_store::` 7 passed · `jwt_lifecycle::` 24 passed · `impersonation_handoff::`
  10 passed · `mcp::auth` 4 passed · `mcp_slice2` 8 passed
- doctests: 77 passed
- `auth_demo --test auth_jwt_api` (the CI-gated backing test for
  `docs/auth-jwt-api.md`): 5 passed
- clippy at CI's two invocations: clean

## Docs

All four locales (en / de / fr / es): `auth-jwt-api.md` gains a "writing a
durable store" section, `security.md` and `api-conventions.md` note the async
surface, and every snippet awaits.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,30 @@ Headline: **a `ViewSet` can finally be scoped to the caller** (#1183) — the
 authenticated identity is available to filter backends, backends apply to every
 action, and `OwnedBy` does the common case in one line.
 
+### Changed
+
+- **`JtiStore` is async** (#1191) — `is_used` / `mark_used` / `approx_size` now
+  return `JtiFuture<'_, T>`, and everything that consults the store is `async`
+  with it: `JwtLifecycle::{verify_token, verify_access, verify_refresh, refresh,
+  refresh_with, revoke, blacklist_jti, is_blacklisted, blacklist_size}`,
+  `mcp::verify_agent_token`, `tenancy::auth_routes::verify_for_tenant`, and
+  `tenancy::admin::redeem_impersonation_handoff`. While the trait was
+  synchronous, a durable multi-instance store could not simply write — it had to
+  be a hot in-memory map with a background flusher, which is eventually
+  consistent, so a revoked `jti` stayed valid on other replicas for the length
+  of the convergence window. Revocation is exactly the operation that should be
+  immediate, and that constraint came from the signature rather than the
+  problem; a Redis- or Postgres-backed store is now one conditional write.
+
+  **Migration:** add `.await` at those call sites. A synchronous
+  implementation stays a one-line wrapper —
+  `fn is_used<'a>(&'a self, jti: &'a str) -> JtiFuture<'a, bool> { Box::pin(async move { … }) }`.
+  The trait returns boxed futures rather than using `async fn`, because every
+  consumer holds it as `Arc<dyn JtiStore>` and native `async fn` in traits is
+  not dyn-compatible. Token **expiry is now checked before** the store is
+  consulted, so an expired token costs no round trip. `InMemoryJtiStore`
+  behaves exactly as before.
+
 ### Fixed
 
 - **ViewSet: 401 for anonymous, 403 for authenticated-but-unauthorised**
--- a/crates/rustango/examples/auth_demo/tests/auth_jwt_api.rs
+++ b/crates/rustango/examples/auth_demo/tests/auth_jwt_api.rs
@@ -15,46 +15,46 @@ fn jwt() -> JwtLifecycle {
     JwtLifecycle::new(b"a-signing-secret-at-least-32-bytes-long!!".to_vec())
 }
 
-#[test]
-fn login_issues_a_pair_and_the_token_types_are_distinct() {
+#[tokio::test]
+async fn login_issues_a_pair_and_the_token_types_are_distinct() {
     let j = jwt();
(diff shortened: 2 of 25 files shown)


ujeenet:
## Review

Checked the thing most likely to be wrong in a sync→async trait conversion: a
`std::sync::Mutex` guard held across a suspend point. It is clean. Every
`Box::pin(async move { … })` body — the real `InMemoryJtiStore` impls, the doc
example, and the `AwaitingStore` test — acquires the lock *after* any `.await`
and never awaits while holding it, so no guard crosses a yield and nothing can
deadlock or block the executor.

The `AwaitingStore` test is the right test to have written: it yields between
the check and the record, which is the only way to prove the single-use
guarantee actually survives concurrency rather than being an artefact of the
in-memory impl never suspending.

Two things worth stating explicitly before merge, neither blocking:

### 1. This is a breaking change to a public trait — confirm the version bump

`JtiStore` gains a new signature and nine public functions become `async`. Any
downstream store implementation stops compiling. The CHANGELOG documents it
under **Changed** with a migration snippet, which is right; just make sure the
release that carries this is a minor bump and the yank/upgrade guidance is
consistent with how 0.51.x handled it.

### 2. `blacklist_size()` became async for an admin-dashboard call

That is the one method where the async conversion buys nothing today — it exists
for status displays, and `InMemoryJtiStore` answers from a map. It is still the
right call (a durable store cannot count synchronously), but it means a
dashboard render now awaits. Worth confirming no dashboard path calls it in a
loop over tenants.

### On the design choice

Hand-rolling `JtiFuture` instead of `#[async_trait]` is correct here and the
reasoning holds up: `async-trait` is an optional dependency enabled only by
`cache`/`storage`/`jobs`/`oauth2`/`email`/`secrets`, while `jti_store` is
ungated — the macro would have broken exactly the minimal feature sets that
#1208 just fixed. Worth keeping that comment in the file; it is the kind of
constraint someone will otherwise "simplify" away.

Issue #1209, opened 2026-08-13
`cargo rustango new --template api` generates a project that cannot compile


ujeenet:
Child of #1207.

## Reproduction

```sh
cargo rustango new probe --template api
cd probe && cargo build
# error: could not compile `rustango` (lib) due to 75 previous errors
```

Reproduced against both the published `rustango 0.51.2` from crates.io (what a
real user gets) and the current working tree, so it is not a release artifact.

## Cause

`crates/cargo-rustango/src/main.rs:126-128` emits, for the `api` template:

```rust
Self::Api => format!(
    r#"{{ version = "{v}", default-features = false, features = ["postgres", "manage"] }}"#
),
```

`postgres,manage` is one of the feature sets that does not build — see the
feature-gate issue filed alongside this one. The unresolved symbols are `tera`
(31), `rand` (15), `tower` (6), `async_trait` (4), `hmac` (1), plus
`AuthBackend` not being dyn-compatible without `async_trait`.

## Impact

One of three shipped templates is unusable. A first-time user following
`docs/scaffolding.md` for a JSON-only service gets 75 compiler errors pointing
into framework internals, with nothing indicating the template chose an
unsupported combination.

## Fix

Fixing the feature gates fixes this template for free. Failing that, the `api`
template must select a combination that actually builds. Either way this needs
the generate-and-check CI test from #1207, or it regresses silently again.

ujeenet:
Fixed on `develop` by #1215 (ca77265).

The dependency no longer pins a backend: `api` takes `["manage"]`, and
fullstack/tenant take a new `batteries` feature (rustango's default set minus
the backend), so the project forwards the backend through its own feature and
`#[derive(Model)]` sees a consistent view.

Verified by `tests/generated_project_compiles.rs`, which generates each
template and compiles it — the api template went from **75 errors to clean**,
and is now checked on every PR by the `scaffolded_projects` CI job.

PR #1212, opened 2026-08-13
fix(features): gate modules on the optional deps they use (#1208)

### LINKED ISSUE #1208: Feature-gate gaps: unconditional modules reference optional deps, so most minimal feature sets fail to build
Child of #1207. This is the **root cause** of the `--template api` failure, and
it reaches much further than the scaffolder.

## The shape of it

Nine `pub mod` declarations in `crates/rustango/src/lib.rs` are **unconditional**
but their bodies reference **optional** dependencies that only exist when some
unrelated feature is enabled:

| Module | lib.rs | needs | gated? |
|---|---|---|---|
| `auth_backends` | 1220 | `async_trait` | no |
| `lorem` | 1155 | `rand`, `tera` | no |
| `random` | 1140 | `rand` | no |
| `messages` | 1210 | `tera`, `tower`, `hmac` | no |
| `urls` | 1135 | `tera` | no |
| `template_extensions` | 1079 | `tera` | no |
| `template_debug` | 1071 | `tera` | no |
| `template_context_processors` | 1075 | `tera` | no |

Two more gates are too narrow for their consumers:

- **`url_codec`** is `#[cfg(any(signed_url, auth_flows, tenancy))]` but is
  imported unconditionally by `admin/audit.rs:461`, `admin/helpers.rs:405`,
  `admin/views.rs:1514` and `pagination.rs:181`. So `admin` **without**
  `tenancy`/`signed_url` cannot build.
- **`list_params`** is `#[cfg(any(tenancy, template_views))]` but is used by
  `viewset/mod.rs:1661,1715`, which is gated on `any(admin, tenancy)`. So
  `admin` alone cannot build.

## Measured impact

```
postgres,manage                                 FAIL   ← the `api` template's exact set
sqlite,manage                                   FAIL
sqlite,manage,template_views                    FAIL
postgres,manage,admin                           FAIL
sqlite,admin                                    FAIL (26 errors)
sqlite,manage,admin                             FAIL (26 errors)
sqlite,admin,signed_url                         FAIL (16 errors)
sqlite,admin,signed_url,template_views          FAIL (14 errors)
sqlite,template_views                           FAIL (61 errors)
sqlite                                          FAIL (151 errors)

sqlite,tenancy                                  OK
sqlite,manage,tenancy                           OK
sqlite,admin,template_views,sessions,signed_url OK
```

Effectively **only feature sets that pull in `tenancy`** (or a fairly heavy
hand-picked combination) build. `tenancy` happens to enable `dep:tower`,
`dep:rand`, `dep:hmac` and `dep:async-trait` and thereby masks the gaps — which
is exactly why the litmus in CI (`--no-default-features --features sqlite,tenancy`)
stays green while nearly every other minimal combination is broken.

This is the same class as the already-known "`--features sqlite` alone fails"
gap; that one is not an isolated case but one instance of a general problem.

## Why it matters beyond the scaffolder

The feature table is a public promise. A user reading `Cargo.toml` reasonably
expects `default-features = false, features = ["sqlite", "admin"]` to work — it
is exactly the "smaller dep tree" the comments advertise. Today it does not, and
the failure is 26 errors deep inside the framework, which reads as "rustango is
broken" rather than "that combination is unsupported".

## Fix directions

1. Gate the eight modules on the features their bodies actually need (`urls`,
   `messages` and the `template_*` trio behind a Tera-bearing feature;
   `auth_backends` behind an `async-trait`-bearing one; `random`/`lorem` behind
   `rand`) — or make the deps non-optional if they are genuinely core.
2. Widen `url_codec` and `list_params` to cover `admin`.
3. Add a CI matrix over the combinations that are *claimed* to work, so the
   promise is enforced rather than assumed. The single `sqlite,tenancy` litmus
   demonstrably does not catch this.

## Reproduction

```sh
cargo check -p rustango --no-default-features --features postgres,manage
cargo check -p rustango --no-default-features --features sqlite,admin
```

ujeenet:
Closes #1208. Unblocks #1209 (the `api` template).

## What was wrong

Nine modules are declared `pub mod` with **no `cfg`**, but their bodies use
**optional** dependencies. They compiled only when some unrelated feature
happened to pull the crate in. Measured on develop:

| feature set | result |
|---|---|
| `postgres,manage` — what `cargo rustango new --template api` emits | **75 errors** |
| `sqlite,template_views` | 61 errors |
| `sqlite,admin` | 26 errors |
| `sqlite,manage,admin` | 26 errors |
| `sqlite,admin,signed_url` | 16 errors |
| `sqlite,manage` | broken |
| `postgres,manage,admin` | broken |
| `sqlite,tenancy` | OK |

Effectively **only combinations including `tenancy` built** — `tenancy` enables
`rand` + `tower` + `hmac` + `async-trait` and masked every gap. The one
combination CI checks, `--no-default-features --features sqlite,tenancy`, is one
of the masking ones. That is why this survived: the promise in the feature table
was never tested anywhere it could fail.

## The fix

`cfg(feature = "...")` cannot ask *"is `dep:rand` enabled?"*, so each optional
dependency a module needs gets an internal capability feature — `_rand`,
`_tera`, `_tower`, `_async_trait`, `_base64`, `_signing`. Product features
enable the capability instead of the raw `dep:`, and each module gates on the
capability it actually needs.

Two modules became **unconditional** instead: `url_codec` and `list_params` are
pure `std` with no dependencies at all, were gated anyway, and are imported by
unconditional modules (`pagination`, `urls`, `crypto`, `default_filters`,
`viewset`) — their gates could never have been satisfiable.

Three call-site bugs of the same family came out with it:

- `manage` merged the `admin`-gated health router unconditionally
- `migrate::runner` and `sql::m2m` emitted `signals` without the feature
- `sql::mod` re-exported `sqlite_connect_options` one gate short of its callers

`url_codec`'s two base64 helpers are gated rather than the whole module, so the
URL/IRI parity surface stays available in every feature set.

## Not fixed here

`--features sqlite` **alone** still fails (50 errors). It is the same class —
`cookies`, `http_methods`, `redirects`, `flatpages`, `auth_decorators`,
`i18n::timezone`, `audit` and `events` need `axum`/`tokio` — but it needs
`_axum`/`_tokio` capabilities across another eight modules and deserves its own
review. Worth noting the manifest advertises it ("Drop `default-features` for
the bare ORM"), so the promise is still outstanding; I'll file it as a follow-up
rather than widen this PR.

## Keeping it fixed

New `feature_combos` CI matrix builds all seven combinations with
`RUSTFLAGS: -D warnings`. Warnings are errors *there specifically* because
cfg'd-out code is exactly what leaves unused imports and unconstructed variants
behind, and that warning is usually the only signal a gate is wrong.

## Verification

| check | result |
|---|---|
| all seven combos | **build clean, zero warnings** |
| `cargo check -p rustango` (defaults) | 0 errors, 0 warnings |
| `cargo check -p rustango --all-features` | 0 errors, 0 warnings |
| `cargo test --workspace --all-features --no-run` | Finished |
| `cargo test -p rustango --all-features --lib` | 3578 passed |
| `cargo test -p rustango --no-default-features --features sqlite,tenancy,testkit --lib` | 2313 passed |
| regenerated `--template api`, built against this tree via `[patch.crates-io]` | **compiles** (was 75 errors) |

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -190,6 +190,9 @@ jobs:
       fail-fast: false
       matrix:
         features:
+          - sqlite                       # the bare ORM the manifest advertises
+          - postgres
+          - mysql
           - postgres,manage              # `cargo rustango new --template api`
           - sqlite,manage
           - sqlite,admin
@@ -206,6 +209,10 @@ jobs:
       - run: cargo check -p rustango --no-default-features --features ${{ matrix.features }}
         env:
           RUSTFLAGS: "-D warnings"
+      # Also run the unit tests: `cargo check` misses `#[cfg(test)]` code, and
+      # that is exactly where a gate mismatch hides — a test module can outlive
+      # the gated item it exercises without the library ever noticing.
+      - run: cargo test -p rustango --no-default-features --features ${{ matrix.features }} --lib
 
   sqlite_litmus:
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -173,6 +173,40 @@ jobs:
   # Runs `cargo check`, `cargo clippy`, and the SQLite pragma
   # regression test (foreign_keys ON, WAL on file-backed, busy_timeout
   # set, FK violation rejected — see tests/sqlite_pragmas_live.rs).
+  # #1208 — the feature table is a public promise, so check the promise.
+  # `sqlite_litmus` below is a single combination, and it happens to be one
+  # that *masks* gate gaps: `tenancy` pulls in rand / tower / hmac /
+  # async-trait, so modules that use those without gating on them still
+  # compiled. Meanwhile `postgres,manage` (what `cargo rustango new
+  # --template api` generates) failed with 75 errors, and `sqlite,admin`
+  # with 26 — undetected, because nothing built them.
+  #
+  # Warnings are errors here, unlike the rest of CI: these combinations are
+  # exactly where cfg'd-out code leaves unused imports and unconstructed
+  # variants behind, and a warning is the only signal that a gate is wrong.
+  feature_combos:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        features:
+          - postgres,manage              # `cargo rustango new --template api`
+          - sqlite,manage
+          - sqlite,admin
+          - sqlite,manage,admin
+          - postgres,manage,admin
+          - sqlite,template_views
+          - sqlite,tenancy               # the pre-existing litmus
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@1.88
+      - uses: Swatinem/rust-cache@v2
+        with:
+          key: feature-combos
+      - run: cargo check -p rustango --no-default-features --features ${{ matrix.features }}
+        env:
+          RUSTFLAGS: "-D warnings"
+
   sqlite_litmus:
     runs-on: ubuntu-latest
     steps:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,32 @@ action, and `OwnedBy` does the common case in one line.
 
 ### Fixed
 
+- **Most minimal feature sets did not build** (#1208) — nine modules were
+  declared `pub mod` with no `cfg` while their bodies used *optional*
+  dependencies (`tera`, `rand`, `tower`, `hmac`, `async-trait`), so they only
+  compiled when some unrelated feature happened to pull the crate in. Measured
+  before: `postgres,manage` **75 errors**, `sqlite,admin` 26, `sqlite,manage`,
+  `postgres,manage,admin`, `sqlite,template_views` all broken — effectively
+  only combinations including `tenancy` worked, because `tenancy` enables all
+  four crates and masked every gap. `--no-default-features --features
+  sqlite,tenancy`, the one combination CI checked, was one of the masking ones.
+
+  `cargo rustango new --template api` emits exactly `postgres,manage`, so a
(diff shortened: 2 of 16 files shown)


ujeenet:
## Review (self-review)

One real finding, and it is mine.

### `tenancy` silently gains a dependency it did not have — should fix

On `develop`, `tenancy` lists `dep:hmac` and `dep:sha2` and **not** `dep:base64`:

```toml
tenancy = [ …, "dep:hmac", …, "dep:sha2", … ]
```

This PR replaces that pair with `_signing`, which is defined as
`["_base64", "dep:hmac", "dep:sha2"]`. So every `tenancy` build now pulls
`base64` where it previously did not.

`base64` is small and almost certainly already in a tenancy dep graph
transitively, so the practical impact is near zero — but it is an undisclosed
widening of a public feature's dependency set, which is exactly the kind of
drift this PR exists to stop. Fixing it my way rather than hand-waving it:
give `tenancy` the precise set it had (`dep:hmac`, `dep:sha2`) and keep
`_signing` for the modules that genuinely need all three (`messages`).

I will push that fix. Flagging it rather than quietly amending because the whole
point of the capability features is that the feature table stops lying about
what it pulls in.

### Things I deliberately did *not* do, for the record

- **`url_codec` / `list_params` became unconditional rather than gated.** Both
  are pure `std`. Gating them on something would have been the symmetric-looking
  change and the wrong one — there is no dependency to gate *on*, and their old
  gates were unsatisfiable for the unconditional modules importing them.
- **Left `--features sqlite` alone broken.** It is the same class but needs
  `_axum`/`_tokio` across eight more modules; it is #1213, stacked on this.
- **Did not touch the 812-warning clippy baseline.** `-D warnings` is scoped to
  the new `feature_combos` job only, where cfg'd-out code makes warnings
  load-bearing signal rather than noise.

PR #1213, opened 2026-08-13
fix(features): make the bare ORM build work (`--features sqlite` alone)

ujeenet:
Stacked on #1212 — review that first; this branch targets it, not `develop`.

## What was wrong

`--no-default-features --features sqlite` failed with **50 errors**, as did bare
`postgres` and `mysql`. The manifest advertises exactly that build:

> Drop `default-features` for the bare ORM (core + query + sql + migrate).

Same root cause as #1208, two more dependency families.

## tokio is no longer optional

`sqlx` is a **hard** dependency built with `runtime-tokio`, so tokio was already
compiled into every build — confirmed with `cargo tree -p rustango
--no-default-features --features sqlite`, which shows `tokio v1.52.1` present and
`axum` absent. `optional = true` only controlled whether rustango was allowed to
*name* the crate it was already linking.

And the modules that need it are core, not opt-in:

- `sql::executor::atomic` — the transaction on-commit registry is *entirely* a
  `tokio::task_local!`
- `audit` — the audit-source task-local
- `events` — the event bus, on `tokio::sync`

Gating those would have silently removed transaction hooks from a bare-ORM
build, which is a worse answer than depending on a crate that is already there.

## axum gets a capability

Unlike tokio, axum genuinely is absent from a bare build, so it gets `_axum`.
`http_methods`, `auth_decorators`, `redirects` and `flatpages` are axum
middleware end to end and gate as modules.

Where a module is mostly dependency-free, only the axum-typed **items** gate, so
the useful surface survives:

| module | gated | still available bare |
|---|---|---|
| `cookies` | `Cookie::header_value` | the whole builder + `build() -> String` |
| `i18n::timezone` | `from_cookie`, `from_request_headers` | offset parsing, activation, task-local |
| `test_assertions` | the `axum::Response` assertions | `query_counter` — the executor bumps it on **every** query, so the module stays unconditional |

## Test modules follow their items

`url_codec`'s base64 cases, `cookies`' `header_value` pair, `timezone`'s four
header tests. Two `manage` tests for the `admin`-gated `try_mount_welcome` were
never gated at all — invisible until `postgres,manage` compiled in the first
place.

The `feature_combos` matrix gains the three bare backends and now runs `--lib`
as well as `check`. That matters: `cargo check` never compiles `#[cfg(test)]`
code, and a test module outliving the gated item it exercises is precisely how
this class of bug hides.

## Verification

Every combination builds clean **and passes its unit tests**:

| features | tests |
|---|---|
| `sqlite` | 1443 |
| `postgres` | 1417 |
| `mysql` | 1449 |
| `postgres,manage` | 1533 |
| `sqlite,manage` | 1559 |
| `sqlite,admin` | 2069 |
| `sqlite,manage,admin` | 2088 |
| `postgres,manage,admin` | 2075 |
| `sqlite,template_views` | 2110 |
| `sqlite,tenancy` | 2313 |

No regressions: `--all-features` still 3578 passing, and `cargo test --workspace
--all-features --no-run` finishes. The 6 remaining warnings in bare-sqlite's
test build (`dateformat` snake-case names, two unused variables) are
pre-existing and appear identically under `--all-features`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -190,6 +190,9 @@ jobs:
       fail-fast: false
       matrix:
         features:
+          - sqlite                       # the bare ORM the manifest advertises
+          - postgres
+          - mysql
           - postgres,manage              # `cargo rustango new --template api`
           - sqlite,manage
           - sqlite,admin
@@ -206,6 +209,10 @@ jobs:
       - run: cargo check -p rustango --no-default-features --features ${{ matrix.features }}
         env:
           RUSTFLAGS: "-D warnings"
+      # Also run the unit tests: `cargo check` misses `#[cfg(test)]` code, and
+      # that is exactly where a gate mismatch hides — a test module can outlive
+      # the gated item it exercises without the library ever noticing.
+      - run: cargo test -p rustango --no-default-features --features ${{ matrix.features }} --lib
 
   sqlite_litmus:
     runs-on: ubuntu-latest
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,8 +33,37 @@ action, and `OwnedBy` does the common case in one line.
   `admin`-gated health router unconditionally, and `migrate::runner` /
   `sql::m2m` emitted `signals` without the feature.
 
-  A `feature_combos` CI matrix now builds all seven combinations with
-  `-D warnings`, so a new gate gap fails the build instead of shipping.
+  A `feature_combos` CI matrix now builds every combination with `-D warnings`
+  **and runs its unit tests**, so a new gate gap fails the build instead of
+  shipping. (`cargo check` alone misses `#[cfg(test)]` code, which is exactly
+  where a gate mismatch hides.)
+
+- **The bare ORM build finally works** — `--no-default-features --features
+  sqlite` (or `postgres`, or `mysql`) failed with **50 errors**, despite the
+  manifest advertising exactly that: *"Drop `default-features` for the bare ORM
+  (core + query + sql + migrate)"*. Same root cause as above, two more
+  dependency families:
+
+  **tokio is no longer optional.** `sqlx` is a hard dependency built with
+  `runtime-tokio`, so tokio was already compiled into every build — `optional =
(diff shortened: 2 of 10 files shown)


ujeenet:
## Review (self-review)

Two findings, both mine, neither behavioural.

### 1. Stale comment above `runtime` — should fix

`crates/rustango/Cargo.toml` still carries this above the feature:

```
# v0.31.1 (#4): `dep:tokio` so the macro can emit
# `#[::rustango::__private_runtime::tokio::main]` and user crates no
# longer need to add tokio to their own Cargo.toml just to use the
# rustango entrypoint shim.
runtime = ["dep:tracing-subscriber", "dep:tracing-appender"]
```

This PR removed `dep:tokio` from that list (tokio is unconditional now), so the
comment documents a line that is no longer there. The *rationale* still matters
— user crates still don't need their own tokio — so the comment should be
rewritten, not deleted.

Same applies to `runserver`, whose comment says "Implies axum + tokio" while the
feature now reads `["_axum"]`.

### 2. Three features are now empty lists

`signals = []`, `sse = []`, `scheduler = []` — all three existed only to pull
`dep:tokio`. They still gate their modules, so they are meaningful and must
stay, but an empty feature list reads like dead config to the next person. Worth
one line each saying "gates the module; the tokio dependency it used to carry is
now unconditional".

### On the load-bearing decision

Making `tokio` non-optional is the part worth scrutiny, so to make the evidence
easy to re-check:

```
$ cargo tree -p rustango --no-default-features --features sqlite -e normal | grep 'tokio v'
│   │   ├── tokio v1.52.1
```

sqlx is a hard dependency built with `runtime-tokio`, so tokio was already
compiled in a build that named no tokio feature at all. `optional = true` was
only controlling whether rustango could *name* the crate it was already linking.

The alternative — gating `sql::executor::atomic`, `audit` and `events` — would
have removed transaction on-commit hooks from the bare-ORM build, i.e. made the
advertised "bare ORM" quietly weaker than the ORM. That seemed clearly worse
than depending on something already present.

The manifest change is still downstream-visible (`tokio` moves optional →
required), so it wants to ride a minor bump, not a patch.

Issue #1223, opened 2026-08-25
Background jobs have no tenant context — the per-tenant worker fan-out is hand-rolled


ujeenet:
## Problem

A worker is not a request, so nothing resolves a tenant for it — and the job
layer offers no way to attach one. Today:

- `Job::run(&self)` ([`jobs/mod.rs`](https://github.com/ujeenet/rustango/blob/develop/crates/rustango/src/jobs/mod.rs#L104))
  receives only the deserialized payload. No pool, no `Org`, no context of any kind.
- `DatabaseJobQueue::with_workers_pool` binds one `sql::Pool` at construction and
  keeps it for the queue's lifetime.
- `docs/jobs.md` had no mention of tenancy at all until the section added
  alongside this issue.

The pattern that *is* safe today — one queue per tenant pool — works because of
where the table lives, not because the job layer enforces anything:

| Mode | Why it holds |
|---|---|
| `database` | `rustango_jobs` sits inside the tenant's own DB / SQLite file; other tenants' rows are not in the table the worker queries. |
| `schema` (PG) | `TenantPools::scoped_pool` bakes `search_path` into the pool's **connect options** rather than issuing a per-request `SET`, so a worker that lives for days stays scoped on every checkout. |

That's sound, but the operator has to assemble it by hand, and the failure mode
of getting it wrong (one queue on the registry pool, `org_id` in the payload,
one `run()` that forgets to re-scope) is a silent cross-tenant write.

## Gaps

1. **No worker fan-out.** Nothing enumerates active tenants and supervises a
   worker set per tenant. Callers hand-roll the loop over
   `Org::objects().where_(Org::active.eq(true))` + `scoped_pool_dyn`.
2. **New tenants get no workers until restart.** A tenant provisioned after boot
   is invisible to a fan-out built once at startup. Needs a refresh path, ideally
   hooked to the same invalidation `TenantPools::invalidate` already drives.
3. **No tenant in the job context.** `run(&self)` can't ask which tenant it's
   running for, so a job that needs the tenant slug (branding, per-tenant config,
   logging) must duplicate it into every payload.
4. **`reclaim_stuck_jobs_pool` is per-pool.** Recovering crashed workers across N
   tenants means N sweeps the caller schedules itself.
5. **Nothing warns about the unsafe shape.** Building a queue directly on
   `registry_pool()` in a tenancy-enabled app is almost always a mistake and
   passes silently.

## Proposed shape

Sketch, not settled:

[...]

PR #1225, opened 2026-08-25
fix(tenancy): reset search_path when a schema-mode TenantConn is released

### LINKED ISSUE #1224: Schema-mode tenancy leaks search_path between registry-pool borrowers — cross-tenant reads from background work
## Problem

Schema-mode tenancy issues a **session-level** `SET search_path` on a connection
borrowed from the **shared registry pool**, and nothing ever resets it. The
connection returns to the pool still pointing at that tenant's schema, so the
next borrower silently inherits it.

Three facts compose:

1. [`TenantPools::acquire`](https://github.com/ujeenet/rustango/blob/develop/crates/rustango/src/tenancy/pools.rs#L717)
   runs `SET search_path TO "<schema>", public` on a connection from the shared
   registry pool.
2. [`TenantConn`](https://github.com/ujeenet/rustango/blob/develop/crates/rustango/src/tenancy/pools.rs#L863)
   has no `Drop` — it just returns the `PoolConnection` to the pool.
3. No pool in the crate installs an `after_release` hook (`grep -rn after_release
   crates/rustango/src` → nothing), and sqlx only calls `ping()` on release
   (`sqlx-core-0.8.6/src/pool/connection.rs:311`). It never issues `DISCARD ALL`
   or `RESET`.

## Reproduction (live PG, `max_connections(1)`)

```
thread 'schema_mode_acquire_resets_search_path_on_release' panicked at
crates/rustango/tests/pools_live.rs:187:5:
assertion `left == right` failed: registry connection inherited the tenant's
search_path after release
  left: "acme_tp_leak"
 right: "public"
```

Acquire for a schema-mode tenant, drop the `TenantConn`, then take a plain
`pool.acquire()` from the same registry pool: `current_schema()` is still the
tenant's.

## Impact

Any query on the registry pool against a table that also exists in tenant
schemas (`rustango_users`, `rustango_audit_log`, `rustango_media`, app models)
resolves against **whichever tenant last used that connection**.

Concretely:

- It makes the contract at
  [`extractors/tenant.rs:148`](https://github.com/ujeenet/rustango/blob/develop/crates/rustango/src/extractors/tenant.rs#L148)
  — *"queries against it would hit `public`"* — untrue. In schema mode
  `Tenant::pool()` is non-deterministically `public` **or another tenant's
  schema**. The note reads as "wrong but harmless"; it is actually
  "intermittently cross-tenant".
- Long-lived background work is the worst amplifier: a worker or
  `scheduler` task handed `registry_pool()` polls for the process lifetime, so it
  *will* draw dirtied connections, and it is the code path with the least tenant
  awareness. The framework's own scheduler-facing helpers are exactly this shape
  — [`MediaLibrary::purge_orphans`](https://github.com/ujeenet/rustango/blob/develop/crates/rustango/src/media/mod.rs#L581)
  selects `rustango_media` rows on one pool and then calls
  `storage.delete(&m.storage_key)`, so a nightly sweep on a dirtied connection
  deletes another tenant's objects.
  [`audit::cleanup_older_than_pool`](https://github.com/ujeenet/rustango/blob/develop/crates/rustango/src/audit.rs#L995)
  and [`prunable::prune_all`](https://github.com/ujeenet/rustango/blob/develop/crates/rustango/src/prunable.rs#L209)
  are the same shape over per-tenant tables.
- It hides in testing. When `public` lacks the tenant table the query *errors* on
  a clean connection and *succeeds against a random tenant* on a dirty one — a
  single-tenant suite passes, production leaks. Nothing in the suite asserts
  connection state after release today.

Note this is a distinct axis from #1223. That issue covers the job layer having
no tenant context and the missing per-tenant worker fan-out, and its "why the
safe pattern holds" table is correct — `scoped_pool` bakes `search_path` into
connect options, so a per-tenant queue stays scoped. This issue is about the
*shared registry pool* leaking session state between unrelated borrowers, which
bites regardless of how the job layer is shaped.

## Fix

`TenantConn` is framework-owned, so the framework can guarantee the reset.
`Drop` cannot be async, so the options are:

- **Reset on release** — `TenantConn::Drop` takes the connection and resets
  `search_path` before it returns to the pool. Complete, and works no matter who
  built the registry pool (which matters: apps and `manage` hand an
  already-constructed `PgPool` to `TenantPools::new`, so `after_release` cannot be
  retrofitted from inside the framework).
- **`after_release` on the pool** — correct and cheaper per request, but only
  reachable for pools the framework constructs.
- Explicit `TenantConn::release().await` — needs every call site to remember, so
  no better than the status quo.

Preference: reset on release inside `TenantConn`, gated on `schema.is_some()` so
it is a no-op for database-mode and for non-PG backends. Plus a hardened
registry-pool constructor that installs `after_release` for the pools the
framework does build.

## Tests

- PG live, `max_connections(1)`: acquire schema-mode tenant → drop → plain
  registry checkout asserts `current_schema() == "public"` (the repro above,
  added to `pools_live.rs`).
- PG live: two schema-mode tenants alternating on a 1-connection pool — tenant B's
  acquire must see B's schema, never A's residue.
- Assert database-mode and non-PG acquires do not emit the reset (no extra
  round-trip on the hot path).

ujeenet:
Fixes #1224.

## The leak

Schema-mode `acquire` issues a **session-level** `SET search_path` on a
connection borrowed from the *shared* registry pool
([`pools.rs:717`](https://github.com/ujeenet/rustango/blob/develop/crates/rustango/src/tenancy/pools.rs#L717)),
and nothing ever undid it. `TenantConn` had no `Drop`, and no pool in the crate
installs an `after_release` hook — sqlx only `ping()`s on release
(`sqlx-core-0.8.6/src/pool/connection.rs:311`), never `DISCARD ALL` / `RESET`.
So the connection went back to the pool still pointing at that tenant's schema
and the next borrower silently inherited it.

Any registry-pool query against a table that also exists in tenant schemas
(`rustango_users`, `rustango_audit_log`, `rustango_media`, app models) therefore
resolved against **whichever tenant last used that connection**.

Reproduced against live PG before touching anything, `max_connections(1)`:

```
assertion `left == right` failed: registry connection inherited the tenant's search_path after release
  left: "acme_tp_leak"    right: "public"
```

And as data — same table name in `public` and in the tenant schema, one row in
the tenant's:

```
assertion `left == right` failed: a released registry connection read the tenant's rows — search_path leaked
  left: 1    right: 0
```

## The fix

`TenantConn` now owns teardown of the session state its acquire path installed:

- `inner` became `Option<PoolConnection<DB>>` so `Drop` can move the connection
  out, plus a `reset: Option<ResetFn<DB>>` fn-pointer set **only** by the
  schema-mode branch of `acquire`.
- `Drop` spawns the reset. It reads as racy but is not: the task **owns** the
  `PoolConnection`, so the pool cannot re-hand it out until the
  `RESET search_path` has landed and the task drops it. No tokio runtime
  available → skip, which is exactly what happened before.
- A failed reset **closes** the connection rather than returning it, so a dirty
  one can never be reused.
- Database-mode and non-PG carry `reset: None` — structurally zero cost, no
  extra round-trip on those paths.

### Why not `after_release`

Cheaper per request, but unreachable here: apps and `manage` hand
`TenantPools::new` an already-built `PgPool`, and sqlx only accepts the hook at
`PoolOptions` construction. Teardown has to live in the type that owns the
checkout.

`Drop` also cannot be async, which rules out doing the reset inline; and an
explicit `TenantConn::release().await` would need every call site to remember,
so it is no better than the status quo.

## Docs

`Tenant::pool()`'s note said queries through it "would hit `public`" — true only
by luck before this change. Reworded to state the guarantee and where it comes
from, since the difference is a wrong-namespace query failing loudly versus
reading another tenant's rows.

## Tests

Two PG live regressions in `pools_live.rs`, both pinned to `max_connections(1)`
so the second checkout is guaranteed to be the same physical connection:

- `schema_mode_acquire_resets_search_path_on_release` — `current_schema()` is
  `public` for a plain registry checkout after the `TenantConn` drops.
- `released_registry_connection_does_not_read_tenant_rows` — the same leak as
  data: a table name present in both namespaces reads 0 rows, not the tenant's 1.

Both were verified to fail with the source change stashed.

## Verification

- `pools_live` 8/8; tenancy PG live suites green: `org_live`, `tenant_auth_live`,
  `tenant_admin_live`, `tenant_migrate_live`, `viewset_tenant_router_live`,
  `branding_live`, `migrate_tenant_storage_live`, `operator_console_orgs_edit_live`.
- Database-mode path (`reset: None`) green on SQLite: `tenant_pools_sqlite_live`,
  `database_pools_sqlite_live`, `database_pools_sqlite_file_live`,
  `pure_sqlite_stack_live`.
- `cargo test --workspace --all-features --no-run` clean.
- `cargo build --no-default-features --features sqlite,tenancy` builds.
- `cargo clippy --features tenancy,postgres --all-targets` reports nothing on the
  new lines.

## Not in scope

#1223 covers the job layer's missing tenant context and worker fan-out — a
separate axis. This is only about the shared registry pool leaking session state
between unrelated borrowers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,42 @@ All notable changes to rustango. The format follows [Keep a Changelog](https://k
 
 ## [Unreleased]
 
+### Fixed
+- **Schema-mode tenancy leaked `search_path` between registry-pool borrowers**
+  (#1224). `TenantPools::acquire` issues a session-level
+  `SET search_path TO <schema>, public` on a connection borrowed from the
+  *shared* registry pool, and nothing undid it: `TenantConn` had no `Drop`, no
+  pool sets an `after_release` hook, and sqlx only pings on release — it never
+  issues `DISCARD ALL` / `RESET`. The connection returned to the pool still
+  pointing at that tenant's schema, so the next borrower silently inherited it,
+  and any registry-pool query against a table that also exists in tenant schemas
+  resolved against whichever tenant used that connection last.
+
+  Long-lived background work is the worst amplifier — a worker holding the
+  registry pool polls for the process lifetime, so it *will* draw dirtied
+  connections. It also hides in testing: when `public` lacks the table, the query
+  errors on a clean connection and succeeds against a random tenant on a dirty
+  one, so a single-tenant suite stays green.
+
+  `TenantConn` now resets `search_path` when released. `Drop` cannot be async, so
+  the reset runs in a spawned task — not racy, because the task owns the
+  `PoolConnection` and the pool cannot re-hand it out until the reset lands. A
+  failed reset closes the connection instead of returning it, and a drop with no
+  runtime available marks it `close_on_drop`, so both failure paths fail closed.
+  `after_release` would have been cheaper per request but is unreachable: apps
+  and `manage` hand `TenantPools::new` an already-built `PgPool`, and sqlx only
+  accepts that hook at `PoolOptions` construction.
+
+  Costs one extra round trip per release, with the connection still checked out
+  while it runs, so a registry pool sized near peak concurrency will feel it as
+  `acquire` latency.
+
+  Also corrects the `Tenant::pool()` docs, stale since v0.38: they claimed
+  schema-mode queries through it hit `public`. The PG extractor resolves that
+  pool via `scoped_pool_dyn`, which builds a *dedicated* pool with `search_path`
+  in its connect options — so `t.pool()` reads the tenant's schema and never
+  borrows from the shared registry pool.
+
 ## [0.52.1] — 2026-08-15
 
 Patch. One fix, no breaking changes — `0.52.0` users can take it directly.
--- a/crates/rustango/src/extractors/tenant.rs
+++ b/crates/rustango/src/extractors/tenant.rs
@@ -80,12 +80,13 @@ pub struct Tenant<DB: Database = DefaultTenantDb> {
     pub org: Org,
     conn: TenantConnCell<DB>,
     /// v0.38 — backend-erasing pool reference for the tenant's storage.
-    /// On PG schema-mode this wraps the registry pool (queries through
-    /// it would hit the `public` schema unless `SET search_path` is
-    /// applied — for that path prefer [`Tenant::conn`]). On non-PG
-    /// (and PG database-mode), this is the tenant's dedicated pool;
-    /// handlers can run `Model::objects().fetch(&t.pool)` for
-    /// tri-dialect ORM queries.
+    /// Always tenant-scoped, by whichever mechanism the storage mode
+    /// calls for: on PG schema-mode [`TenantPools::scoped_pool`] builds
+    /// a dedicated pool with `search_path` in its **connect options**
+    /// (so every checkout is scoped without a per-query `SET`); on
+    /// non-PG and PG database-mode it is the tenant's dedicated pool.
+    /// Handlers can run `Model::objects().fetch(&t.pool)` for
+    /// tri-dialect ORM queries in either mode.
     pool: crate::sql::Pool,
 }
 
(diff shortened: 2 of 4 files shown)


ujeenet:
Review round — three findings, all addressed in f5cffb25.

**1. `Tenant::pool()` doc guarantee was inverted (my error).** The PG extractor resolves `pool` via `scoped_pool_dyn` → `scoped_pool`, which for schema mode builds a *dedicated* pool with `search_path` in its **connect options** — so `t.pool()` reads the tenant's schema and never borrows from the shared registry pool. The pre-existing note claiming it hits `public` had been stale since v0.38, and the first commit here doubled down and called it a guarantee. Both are now corrected, along with the same stale claim on the `pool` field doc and the inline comment at the call site.

This also narrows the exposure described in #1224: the leak is real and the tests prove it, but `Tenant::pool()` is *not* one of its vectors. The exposure is code that uses `registry_pool()` directly — long-lived background workers and the sweeps in #1226.

**2. The SET→TenantConn window was unprotected.** `reset` was only installed after `execute(...).await` returned. If that await is cancelled (request timeout, client disconnect, `select!`) after PG already applied the `SET`, a bare `PoolConnection` dropped — and sqlx-postgres' `ping()` deliberately recovers a connection abandoned mid-statement, so it rejoined the shared pool still tenant-scoped. Same leak, narrower window. The `TenantConn` is now built *before* the statement runs, and the statement goes through it, so any drop on that path resets. A redundant RESET on a connection whose SET failed is harmless.

No test for this one — the window needs a cancellation landing between PG applying the statement and the constructor running, and I could not make that deterministic without an injection point. Flagging it rather than claiming coverage.

**3. Release now costs two round trips** (RESET, then sqlx's ping) with the connection still checked out. Noted in the module doc, which previously described schema mode only as "cheap". Worth keeping in perspective: in schema mode the per-request path already builds a whole new 2-connection pool for `t.pool()` via `scoped_pool`, so one extra RESET is small next to that — but a registry pool sized near peak concurrency will see it as `acquire` latency.

Separately, and out of scope here: `scoped_pool` building a fresh `PgPoolOptions` pool (fresh TCP connect, `max_connections(2)`) on **every** `Tenant` extraction looks like a real per-request cost in schema mode. Happy to file if that is not already known.

Issue #1322, opened 2026-09-10
Operator console: create a tenant, test its connection, and watch provisioning stream live


ujeenet:
Slice 5 of #1317. Depends on #1318, #1319, #1320, #1321.

## Problem

The operator console can list orgs, edit one, re-brand it and impersonate into it. It has **no create route** — an operator who needs a new tenant needs shell access to a box holding registry credentials. This is the slice that closes that.

## What to build

**`GET/POST /operator/orgs/new`** — the create form. Slug, display name, storage mode, backend, database URL, host pattern, port, path prefix. Mode and backend constrain each other: schema-mode is PG-only by language, and the form should say so rather than letting the operator discover it from a validation error.

**`POST /operator/orgs/test-connection`** — a **Test connection** button that runs #1319's probe against the form's URL and renders the taxonomy: not a red X, but "database `acme` does not exist" or "role `app` cannot create tables". No row is written.

**`GET /operator/orgs/provision/{run_id}/stream`** — SSE. A live checklist: each of the #1318 steps as pending / running / ok / failed, each migration named as it is applied, per-step timings, and a terminal result.

The stream must:

- **replay from persistence on connect**, and honour `Last-Event-ID` — see #1321 for why a raw `EventBus` cannot do this
- **terminate** on a terminal event rather than hanging open forever
- render a completed run as a static page — the same URL after the fact is the audit trail

**`GET /operator/orgs/provision/{run_id}`** — the non-SSE view of the same run, for a closed tab or a browser that gave up.

## Authorization

Creating tenants is strictly more dangerous than editing one — it takes a database URL and connects to it. This wants its own operator permission, not a ride on general operator access. Worth checking against [`tenancy/permissions.rs`](../blob/develop/crates/rustango/src/tenancy/permissions.rs) rather than adding a column; see the codename-over-column preference.

## Acceptance

- An operator can go from an empty registry to a migrated, resolving tenant without a terminal.
- A bad URL is caught by **Test connection** and by the pre-flight, and produces no `Org` row.
- The stream survives a mid-run browser reload with no lost steps.
- The stream ends. A finished run does not leave a hanging connection.
- A run that fails renders which step failed and why.
- Live tests on all three dialects, not just PG.

## Watch out for

[...]

Issue #1321, opened 2026-09-10
Persist provisioning runs and their events, so a run survives a reconnect and a second pod


ujeenet:
Slice 4 of #1317. Depends on #1318 and #1320.

## Problem

Slices 1-3 produce a stream of provisioning events. Nothing holds them, and two properties are needed that an in-process bus cannot give:

**Replay.** `crate::sse::EventBus` is a thin `tokio::sync::broadcast` wrapper — a subscriber sees only what is sent *after* it connects. Provisioning starts on a POST and the browser connects on the *next* request, so the opening steps are gone before anyone is listening. A reconnect mid-run loses everything prior.

**Multi-pod.** Behind a load balancer, the operator watching the stream may not be on the pod doing the work. In-process state means that operator sees an empty page while provisioning runs fine somewhere else. This is the same class of problem the resolver already solved with a registry fingerprint rather than a shared cache.

## What to build

Two registry-scope system models — `scope = "registry"` on the derive, the pattern [`OrgHost`](../blob/develop/crates/rustango/src/tenancy/org_host.rs) already uses, so they arrive through the generated system-migration chain:

- **`rustango_provisioning_runs`** — id, slug, requested storage mode / backend / URL (**with the password redacted**), state, requested-by, idempotency key, timestamps, terminal error.
- **`rustango_provisioning_events`** — run id, monotonic sequence, step, status, message, timestamp. The sequence number is what `Last-Event-ID` resumes from.

Plus the state machine: `pending → running → succeeded | failed`, with each of the #1318 steps recorded as it starts and finishes.

## The decision this slice has to settle

**What a failed run leaves behind.** A run that dies at the migration step has a live `Org` row and an unmigrated database. Three options:

1. insert the row `active: false`, flip to `true` only on success
2. roll the row back
3. leave it and expose a resume

Rollback is not always possible — we may have created a schema or a database, and a half-applied migration chain cannot be un-applied. **Option 1 is probably the honest answer**, but it changes what `active` means: today an inactive org is a *suspended customer*, and this would overload it with *never finished provisioning*. `find_active_org_by` filters on exactly this column, so the resolver needs a look before committing. A separate `provisioning_state` column may be cleaner than overloading `active`.

Whatever is chosen, a failed run must not leave something that resolves and serves.

## Acceptance

- A run's full event log replays from persistence, in order, for a subscriber that connects after it started.
- A run started on pod A is fully readable from pod B.
[...]

Issue #1317, opened 2026-09-10
[epic] Provision tenants live from the operator console (and by webhook), with a streamed provisioning console


ujeenet:
## Today

Creating a tenant is a **CLI-only, single-shot, silent** operation.

- `manage create-tenant <slug> --mode database --database-url …` in [`tenancy/manage/tenants.rs`](../blob/develop/crates/rustango/src/tenancy/manage/tenants.rs) is the only path. `manage wizard` wraps it but is still a TTY flow.
- The operator console ([`tenancy/operator_console/mod.rs`](../blob/develop/crates/rustango/src/tenancy/operator_console/mod.rs)) can **list**, **edit**, **re-brand** and **impersonate** an org. There is no create route at all — an operator who needs a new tenant has to get shell access to a box that has the registry credentials.
- For database-mode there is **no connection test**. `create_tenant` takes `--database-url` on faith, `INSERT`s the `Org` row, and only then runs migrations against it. A typo'd host, a wrong password, or a database that doesn't exist yet leaves a **half-provisioned tenant**: a live registry row that resolves, whose migrations never ran. (Schema-mode is better here by accident — it `CREATE SCHEMA`s first so a failed `INSERT` doesn't orphan anything.)
- Migration runs are **opaque**. `migrate_tenants_db` returns a whole `TenantMigrationReport` when it is finished and emits nothing while it works — there is not one `writeln!`, callback or `tracing` span in [`tenancy/migrate.rs`](../blob/develop/crates/rustango/src/tenancy/migrate.rs). On a fresh tenant that is 20-plus migrations of dead air, and if it fails you get one error with no indication of which migration or how far it got.

Onboarding a customer is the single most operationally sensitive thing a multi-tenant framework does, and right now rustango makes it a blind SSH command.

## What this asks for

Provision a tenant from the operator console **or** from an inbound webhook — same engine, two front doors — with a live console view streaming every step.

### 1. A provisioning engine, extracted from the CLI verb

The steps `create_tenant` already performs, factored out of the verb and its `W: Write` so they can be driven by a handler, a webhook, or the CLI:

1. validate slug / host-pattern / storage-mode against the backend (schema-mode is PG-only)
2. **connect-test the target database** and report *what* failed — unreachable, auth, no such database, wrong backend
3. provision (`CREATE SCHEMA` for schema-mode; optionally `CREATE DATABASE` for database-mode)
4. insert the `Org` row
5. run tenant migrations
[...]

Issue #1167, opened 2026-07-21
Bring squash reconciliation to develop (Migration.replaces / --fake / --squash)


ujeenet:
## Background

During the 0.48.0 release we found that `main` had a **direct-to-main** PR (#1159, "Django-style migrations: shipped system migrations + squash reconciliation") that never went through `develop`. Per our workflow (everything lands on `develop`, then promote), `develop` was chosen as canonical for 0.48.0 and `#1159` was superseded/dropped when promoting develop → main.

`#1159` contained **two** capabilities. The framework-shipped *embedded* system migrations approach was intentionally rejected in favor of develop's generate-per-project approach. But **squash reconciliation** is an orthogonal, still-wanted feature that develop currently lacks.

## What to (re)implement on develop

Squash reconciliation via `Migration.replaces`:

- The runner emits plain `CREATE TABLE`, so a squashed `0001` not in the ledger collides with existing tables on an **existing** DB.
- A migration that lists what it `replaces` should **fake-apply** (record it + tombstone its predecessors, **no DDL**) when either:
  - its predecessors are already applied, **or**
  - its tables already exist (cross-ledger `--fake-initial`).
- Runs normally (real DDL) on a fresh DB; **refuses** a partial/inconsistent state.
- `migrate --fake <name>` — operator escape hatch to stamp a migration as applied without running it.
- `migrate --squash` — stamps the generated squash migration's `replaces` list.

## Reference

The original main-only implementation is at commit `7b15a7d` (and the squashed `#1159` = `e8c8dc0`), now off the main lineage after the 0.48.0 promote (`12a9536`). Use it as a reference, but reimplement against develop's **generate-per-project** system-migration design (not the embedded `SYSTEM_MIGRATIONS` approach, which was rejected).

## Tests

Port `migrate_reconcile_sqlite_live` (fresh / same-ledger / cross-ledger / partial-state cases) onto develop's design; verify tri-dialect (SQLite/PG/MySQL).

ujeenet:
[...]

PR #1324, opened 2026-09-10
feat(migrate): report progress while a migration run is happening

### LINKED ISSUE #1320: Migration runners emit nothing while they work — add a progress sink
Slice 3 of #1317. Independently useful; no dependency on #1318.

## Problem

Applying migrations is silent from the outside. There is not one `writeln!`, callback or per-migration span in the runner — the caller blocks and eventually gets a result.

- [`migrate_pool_with_ledger_opts`](../blob/develop/crates/rustango/src/migrate/runner.rs) loops `for mig in pending` and returns `Vec<Migration>` at the end.
- [`tenancy/migrate.rs`](../blob/develop/crates/rustango/src/tenancy/migrate.rs) `migrate_tenants_db` loops over orgs, logs one `info!` **per tenant after that tenant finishes**, and returns a whole `TenantMigrationReport` at the end.

Consequences:

- On a fresh tenant that is 20-plus migrations of dead air, then either "done" or one error.
- A failure says which tenant, not which migration, and not how many had already applied.
- Nothing can render progress: not `manage migrate`, not the operator console, not CI.

## What to build

A progress sink threaded through both loops. Events at minimum:

- migration started (name, index, total pending)
- migration applied (name, duration)
- migration **faked** — `migrate_pool_with_ledger_fake_initial` records a create-only migration without running its SQL (#1167 reconciliation); "applied" and "faked" must not look the same to a watcher
- migration failed (name, error)
- per-tenant start / finish, in the tenancy layer

## Keep the public API intact

`migrate_pool`, `migrate_pool_with_ledger` and `migrate_pool_with_ledger_fake_initial` all funnel into the private `migrate_pool_with_ledger_opts`. Add the sink there and give the three public wrappers a no-op — the surgical change is one private signature, and nothing downstream breaks. Same shape in the tenancy layer: `migrate_tenants_db` / `migrate_tenants` / `migrate_tenants_dyn` keep their signatures, with `*_with_progress` siblings.

## Immediate payoff, before any console exists

- `manage migrate` and `manage create-tenant` can print per-migration progress instead of hanging.
- A migration failure names the migration.
- #1223 (background jobs have no tenant context) wants the same per-tenant observability.

## Acceptance

- Existing runner and tenancy migration tests pass **unmodified** — the no-op path must be exactly today's behaviour.
- A test asserts the event sequence for a multi-migration apply, including ordering and the applied-vs-faked distinction.
- `manage migrate` prints per-migration lines.

## Watch out for

**The sink is called from inside `with_migrate_lock_pool`.** That is an advisory lock held for the whole apply. A sink that blocks — a synchronous HTTP post, an unbuffered channel with no receiver — extends the lock hold and can stall every other pod trying to migrate. Make the sink non-blocking by contract and say so in the docs; a broadcast send that drops on a full buffer is the right default.

**Timing is per-migration, not cumulative.** A 40-migration chain where one migration takes 90 seconds is the case this exists to make visible.

ujeenet:
Closes #1320. Slice 3 of #1317 — the one with no dependencies, and the one that pays off on its own.

## Before

Applying migrations was silent. The runner took the migrate lock, applied every pending file, and returned a `Vec<Migration>` when it was done — not one `writeln!`, callback or per-migration span in it. `migrate_tenants_db` logged one line *per tenant, after that tenant finished*.

So: a fresh tenant is twenty-plus migrations of dead air, and a failure tells you which **tenant** died but never which migration, or how many had already landed.

## After

```
migrating 2 tenant(s)…
[1/2] acme
        applied system/0001_rustango_tenant_initial (0.1s)
        applied app/0001_initial (0.0s)
        applied app/0002_add_bio (0.0s)
[2/2] globex
        faked system/0001_rustango_tenant_initial (0.0s)
        ✗ app/0002_add_bio: relation "bio" already exists
ran tenant migrations against 2 tenant(s); 1 failure(s)
  ✓ acme: 3 migration(s)
  ✗ globex: …
```

`migrate::progress` adds an observer — pending count up front, `Started` / `Finished` per migration with its duration, or `Failed` naming the one that died. Every pre-existing entry point passes no observer and behaves exactly as before.

## Three things worth review attention

**`Faked` is not `Ran`.** The runner has three apply paths: run the DDL, run it skipping tables that already exist (`RunPartial`), or record it in the ledger *without running anything* (`--fake-initial` reconciliation, #1167/#1174). Collapsing those into "applied" is how an operator comes to believe DDL ran that never did, so `Outcome` keeps them apart and the CLI prints `faked`.

**Both of a tenant's chains report, tagged.** A tenant migrates the framework's `system/migrations/` chain and then the project's own, into separate ledgers, numbered independently — so **`0001_initial` legitimately lands twice in one tenant's run**. `Chain` says which, rendered `system/…` vs `app/…`. Without it the second one reads as a migration that ran twice.

**Observers must not block, and the docs say why.** Events are emitted from inside the migrate lock (`pg_advisory_lock` / `GET_LOCK`), so a slow observer stalls every other pod trying to migrate. The CLI's own observer writes to the verb's writer, which is the one legitimate exception — one process, one operator, and a stall delays only their own migration. That reasoning is written down where it applies rather than left implicit.

## Tri-dialect, not Postgres-shaped

Both apply loops are instrumented: the tri-dialect `migrate_pool_with_ledger_opts` (on `crate::sql::Pool`) and the PG-typed `migrate_with_ledger` — the latter separately, because it fires the `pre_migrate` / `post_migrate` signals and uses the PG-typed ledger helpers, so it cannot simply be routed through the other. Both tenant fan-outs (`migrate_tenants_db`, and PG's `migrate_tenants` which also handles schema-mode) report too.

Nothing new is gated on `postgres` except the two shims that take a `&PgPool` by signature (`migrate_with_progress`, `migrate_tenants_with_progress`) — mirroring the existing `migrate` / `migrate_tenants` gating.

## API

Additive only; no existing signature changed.

| new | notes |
|---|---|
| `migrate::progress` (`MigrationEvent`, `Outcome`, `MigrationObserver`) | ungated; `MigrationObserver` is blanket-impl'd for `Fn(MigrationEvent)` so a closure works |
| `migrate_pool_with_progress` | ungated, tri-dialect |
| `migrate_pool_with_ledger_fake_initial_with_progress` | ungated; the path where `Faked` actually appears |
| `migrate_with_progress` | `#[cfg(postgres)]` — takes `&PgPool` |
| `tenancy::{Chain, TenantMigrationEvent, TenantMigrationObserver}` | ungated |
| `migrate_tenants_db_with_progress` | ungated, tri-dialect |
| `migrate_tenants_with_progress` | `#[cfg(postgres)]` |

## Verification

- **13 new tests against live Postgres 16, MySQL 8 and SQLite** — one file, shared scenario bodies, three cfg-gated backend modules (the `django6_*` pattern). PG reads `DATABASE_URL`, MySQL `MYSQL_TEST_URL`, both skip when unset.
- `manage_live` 13/13 against PG, including a **new assertion that the verb really emits per-migration lines** — the acceptance criterion, not just the plumbing.
- The five pre-existing migration suites (19 tests) pass **untouched** — that is the proof the silent path is unchanged. One test asserts it directly: same migration dir, two fresh databases, silent vs observed, identical result.
- 3629 lib tests.
- Zero new warnings on `sqlite` / `postgres` / `mysql` / each `+tenancy` / `+signals` / `+testkit` / `--all-features --tests`, plus `--workspace --all-features --no-run` and `cargo fmt --check`.

## What this unblocks

#1321 (persist provisioning runs) and #1322 (the streamed operator console) both need something to stream; this is it. #1223 wants the same per-tenant observability.

### DIFF
--- a/crates/rustango/src/migrate/mod.rs
+++ b/crates/rustango/src/migrate/mod.rs
@@ -34,6 +34,9 @@ pub mod make;
 // queries PG-specific `information_schema` and stays
 // `#[cfg(feature = "postgres")]`-gated inside `manage::run_with_writer`.
 pub mod manage;
+/// Watching a migration run while it happens — the observer the
+/// progress-reporting entry points take.
+pub mod progress;
 mod runner;
 pub mod scaffold;
 pub mod snapshot;
@@ -52,9 +55,12 @@ pub use make::{
 };
 #[cfg(feature = "postgres")]
 pub use manage::{append_data_op, make_data_migration};
+pub use progress::{MigrationEvent, MigrationObserver, Outcome};
 pub use runner::ensure_ledger_pool_with_ledger;
 pub use runner::migrate_pool_with_ledger;
 pub use runner::migrate_pool_with_ledger_fake_initial;
+pub use runner::migrate_pool_with_ledger_fake_initial_with_progress;
+pub use runner::migrate_pool_with_progress;
 // Always-on: tri-dialect entry points (work on PG / MySQL / SQLite via
 // the `Pool` enum), plus the inventory + builder surface.
 pub use runner::{
@@ -67,6 +73,6 @@ pub use runner::{
 #[cfg(feature = "postgres")]
 pub use runner::{
     applied_set, apply_all, downgrade, drop_all, ensure_ledger, migrate, migrate_dry_run,
-    migrate_embedded, migrate_to, unapply, unapply_force,
+    migrate_embedded, migrate_to, migrate_with_progress, unapply, unapply_force,
 };
 pub use snapshot::{FieldSnapshot, IndexSnapshot, RelationSnapshot, SchemaSnapshot, TableSnapshot};
--- a/crates/rustango/src/migrate/progress.rs
+++ b/crates/rustango/src/migrate/progress.rs
@@ -1,0 +1,160 @@
+//! Watching a migration run while it happens.
+//!
+//! The runner used to be silent: it took the migrate lock, applied every
+//! pending file, and returned a `Vec<Migration>` when it was done. On a
+//! fresh tenant that is twenty-plus migrations of nothing, and a failure
+//! reported only which *tenant* died, never which migration or how many
+//! had already landed.
+//!
+//! An observer fixes that without changing what the runner does. Pass one
+//! to [`migrate_pool_with_progress`](super::migrate_pool_with_progress)
+//! and it is called as each migration starts and finishes; pass nothing
+//! and the behaviour is exactly what it was.
+//!
+//! ## Observers must not block
+//!
+//! Events are emitted from **inside the migrate lock** — a PG advisory
+//! lock, a MySQL `GET_LOCK`, held for the whole run. Every other process
+//! trying to migrate is waiting behind it. An observer that blocks, does
+//! synchronous I/O, or sends on a full unbuffered channel extends that
+//! hold for everyone.
+//!
+//! Send on a bounded channel that drops when full, or push onto a queue a
+//! separate task drains. Losing a progress event is a cosmetic problem;
+//! stalling every migrating pod is not.
+//!
+//! ```ignore
+//! use rustango::migrate::{migrate_pool_with_progress, MigrationEvent};
+//!
+//! migrate_pool_with_progress(&pool, dir, &|event: MigrationEvent| {
+//!     if let MigrationEvent::Finished { name, index, total, .. } = &event {
+//!         println!("[{index}/{total}] {name}");
+//!     }
+//! })
+//! .await?;
+//! ```
+
+use std::time::Duration;
+
+/// Something that happened during a migration run.
+///
+/// Owned rather than borrowed on purpose: an observer's whole job is
+/// usually to hand the event to somewhere else — a channel, a broadcast
+/// bus, a run log — and a borrow would tie it to the runner's stack. One
+/// allocation per migration is not measurable against the DDL it
+/// describes.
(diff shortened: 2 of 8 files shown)


ujeenet:
Re-targeted from `develop` to **`release/v0.57.0`**, so this stack now lands on the release branch instead of straight onto develop. Nothing in this PR changed — same commit, same eight files; only the base moved.

The release branch is `develop` plus one commit: the 0.57.0 version bump. Minor rather than patch, because the stack adds `--backend` / `--features` / the wizard to the scaffolder, `manage menu`, and the tenancy verbs that close the console/CLI gap.

The bump moves the workspace version, the three internal path-dep pins, the `rustango-macros` pin inside `crates/rustango` (not a workspace dep, and `cargo metadata` refuses to resolve until it moves too), the renamed-smoke crate, and all eight example lockfiles.

`CHANGELOG.md` is deliberately untouched there. Its `[Unreleased]` section is where the PRs above are adding entries — #1353 in particular — so renaming the heading now would conflict with them. It becomes `[0.57.0]` once the last PR merges and the release is cut.

Merging this will produce a merge commit rather than a fast-forward, since the bump sits beside the stack rather than under it. That is the normal shape for a release branch and needs nothing from you.

The rest of the stack chains off this one unchanged:

```
release/v0.57.0
 └ #1324 → #1325 → #1326 → #1327 → #1328 → #1329
     └ #1346 → #1347 → #1348 → #1349 → #1350 → #1351 → #1352 → #1353
```

Issue #1345, opened 2026-09-12
Interactive mode for the CLI: numbered menus over manage, plus a project-creation wizard with feature selection


ujeenet:
Running `manage` bare should open an interactive menu rather than printing a wall of verbs. Today you have to know the verb, its flags, and their spelling before you can do anything; `manage --help` lists ~37 verbs across tenancy, users, roles, agents and migrations.

There is already the right foundation to extend rather than replace:
- `manage wizard` (`tenancy/manage/wizard.rs`) — a conversational flow that reads from a `BufRead` and writes to the dispatcher's writer, calling internal verb functions directly. No process spawning, no argv reconstruction.
- `manage_interactive::ask` / `ask_password` — TTY-gated prompts already used by `create-tenant`, `drop-tenant`, `purge-tenant` and others for their confirmations.

## Design

**Entry.** `manage` with no argv, and only when stdin is a TTY — piped or CI invocations must keep printing help and exiting non-zero, or a deploy script hangs on a prompt forever. `manage menu` forces it.

**Top level — grouped, not 37 flat items:**

```
rustango manage — acme_dev (postgres, 4 tenants)

  1  Tenants        create, migrate, decommission, connection
  2  Operators      create, deactivate, reset a password
  3  Users & access roles, permissions, superusers
  4  Migrations     registry, tenants, storage mode
  5  Inspect        runs, audit log, list tenants
  6  Setup wizard   first-run: registry + operator + first tenant

  q  quit
>
```

**Second level** lists that group's verbs with one-line descriptions, then prompts for each argument the verb needs, with defaults in brackets and validation reusing the engines' own rules (`validate_slug` and friends) so the menu cannot produce something the verb rejects.

**Rules that keep it honest:**
- Every leaf calls the same internal verb function the argv path calls. The menu is a front end, never a second implementation — same shape as the wizard today.
- Destructive verbs keep their existing typed-confirmation (`purge-tenant` asks for the slug verbatim); the menu does not soften it.
- After running a verb, print the equivalent command line — `→ manage create-tenant acme --mode database --database-url …` — so the menu teaches the CLI instead of hiding it, and the user can paste it into a script.
- `BufRead` + `Write` throughout, so the whole thing is testable with canned input the way `wizard_cmd` already is.
- Ctrl-C / EOF at any prompt aborts without writing.

[...]

PR #1347, opened 2026-09-12
feat(tenancy): `manage menu` — numbered choices over the verbs

ujeenet:
Part of #1345. Stacked on #1345's scaffolder slice.

There are 42 tenancy verbs. `--help` tells you they exist; it does not help you run one you have never run before, and the flags each takes are only discoverable by reading 300 lines of help text.

`cargo run -- menu` lists them by group, numbered:

```
  rustango manage — pick an action

  TENANTS
     1) list-tenants             every tenant in the registry
     2) create-tenant            provision a new tenant
     ...
  OPERATORS AND USERS
     6) create-operator          an apex-level account for the operator console
     ...
     q  quit
     ?  every other verb: cargo run -- --help
```

Pick one and it asks for the options that verb will **not** ask for itself, echoes the command line it is about to run, runs it, and comes back for the next action.

### Why the catalog is small

Most verbs already prompt for their own required values on a terminal — `create-tenant` asks for a slug, `create-operator` asks for a username and reads the password without echoing it. Re-asking here would duplicate that and, for passwords, do it worse. So an entry carries only what the verb won't prompt for: `--days`, `--mode`, `--generate`, `--superuser`.

### Why it can't drift

Each pick re-enters the dispatcher's own `match`, which is why the big match moved into `dispatch` and `run_with_writer_and_init` became a one-line wrapper. The menu is a front end, not a second implementation.

The failure mode that shape still allows is a catalog naming a verb the dispatcher dropped — it would fall through to the single-tenant runner and fail with something unrelated. A test reads the match arms and asserts every menu verb is among them.

### DIFF
--- a/crates/rustango/src/tenancy/manage/menu.rs
+++ b/crates/rustango/src/tenancy/manage/menu.rs
@@ -1,0 +1,602 @@
+//! `manage menu` — numbered choices over the tenancy verbs.
+//!
+//! There are 42 verbs. Reading `--help` tells you they exist; it does not
+//! help you run one you have never run before. The menu lists them by group,
+//! asks only for what the verb will not ask for itself, echoes the command it
+//! is about to run, and comes back for the next action.
+//!
+//! ## Why the catalog is small
+//!
+//! Most verbs already prompt for their own required values on a terminal —
+//! `create-tenant` asks for a slug, `create-operator` asks for a username and
+//! reads the password without echoing it (`manage_interactive`). Re-asking
+//! here would duplicate that and, for passwords, do it worse. So [`Action`]
+//! carries only the options a verb will *not* prompt for: `--days`,
+//! `--mode`, `--generate`. Everything else is left to the verb.
+//!
+//! Each pick re-enters [`super::dispatch`], so the menu cannot drift from
+//! what the flags do — same code path, different front end.
+
+use std::io::{BufRead, Write};
+use std::path::Path;
+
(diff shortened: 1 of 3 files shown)


ujeenet:
Hand-tested this on a real terminal and found it deadlocks on most of its own entries — filed as #1360.

Picking a verb that prompts for a value prints the first prompt, consumes the answer, then hangs forever:

```console
$ cargo run -- menu
  > 4
  Same thing without the menu:
    cargo run -- drop-tenant

Tenant slug to drop: nosuchtenant-qa
<nothing, 20s+, still running>
```

The same verb standalone works fine, so it is the nesting: `menu_cmd` holds `stdin.lock()` for the life of the loop while the verb calls `io::stdin().read_line()` underneath it.

Entries needing no input (`list-tenants`, `list-operators`, `list-runs`) execute correctly and return to `Another action?`, which is why the unit tests are green — they drive `menu_cmd` with a `Cursor`, so the lock is never taken, and `picking_an_action_runs_it` happens to pick `list-tenants`.

This is the path the PR description calls the design:

> asks only for what the verb will not ask for itself

…so it is the majority of the menu, not an edge case. I flagged the risk in the module docs while writing it, said I would check it against a real terminal, and did not.

Worth holding this PR until it is fixed rather than landing a menu that hangs on `create-tenant`. Happy to push the fix here — dropping the lock around `dispatch`, or threading the reader through the way `wizard.rs` already does.

Issue #1344, opened 2026-09-12
CLI should mirror every action the operator console can do


ujeenet:
The console and the CLI have drifted. Work that started CLI-only grew a web UI, and work that started in the console never got a verb — so neither surface is the complete one, and which half you get depends on which you reach for.

**In the console, not on the command line:**
- bind / unbind / park an extra hostname (`org_host` has the API, no verb)
- create an operator interactively, deactivate one, reset another's password (`create-operator` and `reset-operator-password` exist, but there is no `deactivate-operator` / `list-operators` with state)
- read the audit log (`/audit` renders it; nothing prints it)
- list provisioning and migration runs, or replay one (`/orgs/provision`)
- test an existing tenant's connection (`test-tenant-connection` takes a URL, not a slug)
- edit a tenant's `host_pattern`, `port`, `path_prefix`, branding

**On the command line, not in the console** — tracked in #1341: `prewarm-pools`, `migrate-tenant-storage`.

Parity matters beyond tidiness: the CLI is what runs in CI, in a deploy hook, in a cron, and on a box where nobody can open a browser. An action that exists only in the console cannot be automated, and one that exists only in the CLI cannot be delegated to someone without shell access.

Both surfaces already sit on shared engines — `provision`, `decommission`, `migrate_run`, `org_host` — so most of these are argv parsing and rendering over a call that exists. Where they do not (tenant edit), the extraction is the same shape as #1318 and #1340.

Suggested order, by what is most painful to lack:
1. hostnames (`add-host` / `remove-host` / `set-host-enabled` / `list-hosts`)
2. operator management (`deactivate-operator`, `list-operators`)
3. read-only inspection (`list-runs`, `show-run`, `audit-log`)
4. tenant edit

Worth adding a test that asserts the two surfaces stay in step, so this does not silently re-open.

ujeenet:
All four items are on `feat/provisioning-webhook` (unpushed).

**1. Hostnames** — `list-hosts`, `add-host`, `remove-host`, `set-host-enabled --on|--off`, over `tenancy::org_host`. A test writes through the verb and reads through the engine, and the other way round, so "same engine" is asserted rather than asserted-in-a-comment. No `rename-host`: `org_host::generation` fingerprints the table by row count, enabled count, max id and enabled-id sum, and an in-place rename moves none of them, so other pods would keep routing the old name until the TTL expired. The module says so where someone would go to add it.

[...]

Issue #1341, opened 2026-09-12
Operator console: the remaining CLI-only operations (re-test a tenant's connection, prewarm, storage move, audit retention)


ujeenet:
Audit of every registry-scoped capability against the console's route table. These four remain command-line only. Each is small; grouped because they are the tail of the same audit.

**Re-test an existing tenant's connection.** `/orgs/test-connection` is mounted but only reachable from the *create* form. A tenant whose database moved, or whose credentials rotated, cannot be probed from its own edit page — the operator has to start creating a tenant to borrow the button.

**`prewarm-pools`.** Useful after a deploy or a registry restart; no console path.

**`migrate-tenant-storage`.** Moving a tenant between schema and database mode (Postgres). Consequential enough that a console flow with a streamed run would be better than a shell command, and it fits the run/stream machinery.

**`audit-cleanup`.** `cleanup_older_than` / `cleanup_keep_last_n` exist. With `/audit` now readable (347a5708 and earlier), retention becomes visible as a gap — the table grows with every console action and nothing trims it.

Excluded deliberately: tenant-scoped verbs (`create-user`, roles, permissions, API keys, agents, skills) belong to the tenant admin, which impersonation already reaches.

ujeenet:
First of the four shipped in ffb290a3: **re-test an existing tenant's connection**.

`POST /orgs/{slug}/test-connection` resolves the URL server-side from the slug (through the secrets resolver, so vault-backed tenants work) and returns the same fragment the create form uses. The URL never reaches the browser — a test asserts the page offering the button doesn't contain it. Schema-mode tenants answer that there's nothing separate to test rather than failing on a connection they don't have.

Verified live against a tenant pointed at a missing database: *"the server is reachable but has no such database — create it, or fix the name in the URL (tried postgres://rustango:\*\*\*@localhost:5432/moved_away)"*.

Still open here:
- `prewarm-pools`
- `migrate-tenant-storage` (consequential enough to deserve the run/stream treatment, like #1339)
- `audit-cleanup` — worth noting this one is bigger than it looks: `cleanup_older_than` / `cleanup_keep_last_n` are `PgPool`-typed, so exposing retention means making them tri-dialect first. More pressing now that `/audit` is readable and the table grows with every console action.

ujeenet:
Three of the four are done; the fourth is the one I am stopping short of, deliberately.

[...]

PR #1353, opened 2026-09-12
docs: the tenancy CLI in four locales, plus the six fixes QA found

ujeenet:
Stacked on #1352. Documents everything in the stack below it.

The tenancy section of `manage.md` documented fifteen verbs; the dispatcher has more than forty. Everything added while closing #1344, #1345 and #1341 was undocumented — and so were several verbs that predate them (`wizard`, `test-tenant-connection`, `prewarm-pools`).

### `manage.md`, in `en` / `de` / `fr` / `es`

| Added | |
|---|---|
| `menu` / `actions` | the numbered verb menu |
| `wizard` / `init` | the first-run walk-through |
| `edit-tenant` | routing and display config, with the `--clear` distinction |
| `test-tenant-connection` | probe a URL before committing to it |
| **Hostnames** | `list-hosts` / `add-host` / `remove-host` / `set-host-enabled` |
| `list-operators` | who can sign in |
| `set-operator-active` | and the two lockout rules, and why the direction is required |
| **Inspecting what happened** | `list-runs` / `show-run` / `audit-log` |
| `prewarm-pools` | open every tenant's pool up front |

…plus the `--registry` scope on `audit-cleanup`, which now sweeps the registry's own log by default.

### `scaffolding.md`

`--backend`, `--features` with the opt-in table, and the wizard — including that it prints the equivalent command line, which is the reason it is safe to use and then graduate from.

### How the examples were checked

Not written from the source. Every example was run against a live registry, and the prose matches what came back:

```
$ cargo run -- list-operators
username                         active   created_at
----------------------------------------------------------------
admin                            true     2026-09-11 12:42:06Z

1 operator(s), 1 active

$ cargo run -- set-operator-active admin --off
Error: this is the last active operator. Deactivating it would lock everyone
out of the console, and only a shell on the registry could undo it.

$ cargo run -- edit-tenant acme --host-pattern shop.test:8443
Error: host pattern `shop.test:8443` carries a port — the `Host` header is
matched with the port stripped, so this could never match.
```

The normalization claim ("the verbs echo what was **stored**, not what you typed") was checked the same way: `add-host acme "  SHOP.Example.COM  "` prints `bound shop.example.com -> acme`. Test data created along the way was reverted afterwards.

Each documented verb was also checked against the dispatcher's `match` arms, so the docs cannot name a verb that does not exist.

### Translations

Not machine-copied English. Each follows the register its own file already uses — `manage.md` is formal *Sie* in German while `scaffolding.md` is informal *du*; French and Spanish follow their existing tenant/registry terminology — and they were verified in lockstep: identical code-fence counts across all four files, and every new verb present in all four.

`docs_links` passes, so links and anchors resolve in every locale.

### Not changed

`docs/index.toml` still declares `version = "0.53"` against a `0.56.1` workspace. It has been bumped in docs PRs before, but it seeds the docs site's version label, so I left it rather than guess at the right moment for that.


ujeenet:
Pushed a second commit: `--backend` in getting-started, the CHANGELOG entries for the stack, and the broken locale TOCs.

**getting-started** already had a "Choosing a database" section with a SQLite path — documenting the *old* manual route (build with `--no-default-features --features sqlite`, then hand-edit `DATABASE_URL`). `--backend sqlite` does both at generation time, so that is the lead now, with the manual route kept for a project that already exists. My first draft added a paragraph at Step 2 that duplicated that section; it is now one line about the wizard instead.

**CHANGELOG** gains the `[Unreleased]` entries for #1345, #1344 and #1341, and is reordered to Added / Changed / Fixed — it had two `### Fixed` blocks.

**The locale TOCs were broken.** Every translated `manage.md` and `scaffolding.md` translated its headings but kept the English anchors, so all nine TOC entries per file pointed at ids that do not exist there:

```markdown
- [Commandes système](#system-commands)     <- heading is `## Commandes système`
```

`docs_links` cannot catch this — it skips `#`-anchors by design and resolves file paths only. Checking them took a separate pass, and the slug rule it implies is confirmed by the English pages: 0 unresolved out of 517.

That pass also turned up **886 more of the same class across 91 other translated pages** — `security.md`, `admin.md`, `serializers.md`, `viewsets.md`, `orm.md` in all three locales. Filed as #1354 rather than swept in here; a mechanical change that size wants its own review, and the durable fix is extending `docs_links.rs` to check anchors so the next translated page cannot repeat it.

ujeenet:
Third commit here fixes everything the hand-testing pass turned up — #1355, #1356, #1357, #1358, #1359, #1360.

**Placement, up front:** these are code fixes and they landed on the docs branch. I committed them here and pushed before splitting them onto their own branch, and I am not rewriting a pushed branch to tidy that. So this PR is now docs **plus** fixes; the title says so. If you would rather review them apart, say the word and I will revert the commit here and re-apply it as a separate PR stacked on top — that is additive, not a rewrite.

### What they fix

**#1360 — the menu deadlocked on most of its own entries.** `menu_cmd` held `stdin.lock()` for the life of the loop while each verb reached for stdin itself, so picking anything that prompts (`create-tenant`, `drop-tenant`, `list-hosts`, a dozen more) printed the first prompt, took the answer, and hung forever. The wizard had the same shape. Both now take a `LineSource`; the dispatcher passes `SharedStdin`, which locks per read and leaves the stream free between prompts. A `BufRead` is still a `LineSource`, so the `Cursor` tests are unchanged.

**#1355 — contradictory flags resolved last-wins.** `edit-tenant --activate --deactivate` took a tenant offline, printed `updated`, exit 0. `--enabled TRUE` parked a live host, because anything outside a lowercase allow-list read as `false`. Both refused now; `--enabled` parses a closed set case-insensitively.

**#1356 — errors named the wrong object.** `list-hosts nosuchtenant` said "no such host"; one `HostError::NotFound` covered both a missing org and a missing host row. Split into `NoSuchOrg`. An invalid `--kind`/`--state` returned "(no runs)" and is now validated.

**#1357** — `menu` requires a terminal. **#1358** — `build`/`deps`/`examples`/`incremental` refused before any file is written. **#1359** — `.env.example` no longer ships a session secret the framework discards, and `from_env_or_disk` warns when one is set but unusable.

### Testing

A regression per fix, including the weak assertion that let #1356 through (`an_unknown_tenant_is_named` checked only that *something* was said). The menu test asserts control returns and the loop continues — what a `Cursor` can observe; the nested prompt itself needs a terminal and the comment says so rather than implying more coverage than exists.

Verified against a live Postgres registry and a real pty, `--all-features --no-run` clean, 2405 lib tests green.

Issue #1450, opened 2026-09-14
SqlValue::Null binds as text, so writing NULL into any non-text column fails on Postgres


ujeenet:
`SqlValue::Null` is bound as `None::<String>` ([`sql/executor/mod.rs:741`](https://github.com/ujeenet/rustango/blob/develop/crates/rustango/src/sql/executor/mod.rs#L741), and again at `:834` and `:881`):

```rust
SqlValue::Null => $q.bind(None::<String>),
```

That sends the parameter to PostgreSQL with the **text** type OID. PostgreSQL then refuses to assign it to a column of any other type:

```
code: "42804"
message: "column \"uploaded_by_id\" is of type bigint but expression is of type text"
hint: "You will need to rewrite or cast the expression."
```

SQLite and MySQL accept it — they do not type parameters this strictly — so the failure is **Postgres-only**, which is why it has survived.

## Reproduced

Against a real Postgres 16 and MinIO, running the media live suites for the first time (see #1437 — they had never executed):

```
media_live                    6 tests, 6 failed
media_collections_tags_live  16 tests, 16 failed
```

Every failure is this error, on `uploaded_by_id` or `derived_from_id`. The Postgres insert path in `media/mod.rs:1204` builds the row with

```rust
r.uploaded_by_id.map(SqlValue::I64).unwrap_or(SqlValue::Null)
```

so a media upload with no uploader — which is the ordinary case for an anonymous or system upload — cannot be written at all. **Media upload is broken on Postgres**, and the tests that would have said so were reporting green without running.

Note the MySQL branch fourteen lines below binds `r.uploaded_by_id` (an `Option<i64>`) directly and is unaffected. The two branches of the same function disagree, and only the typed one is correct.

## Scope

This is not a media bug. `SqlValue::Null` is constructed at **52 sites across 12+ files** — `soft_delete`, `audit`, `fixtures`, `forms`, `viewset`, `admin/views`, `admin/inlines`, `tenancy/org_edit`, `operator_console`, `migrate/manage`, `template_views`, `core/validate`. Any of them that writes NULL into a non-text column on Postgres hits this. I have confirmed the media path by execution; the others are the same expression and have not been individually exercised.

## The fix is not a one-liner

`None::<String>` has to become a NULL whose type Postgres infers. Options, none free:

[...]

Issue #1449, opened 2026-09-14
A password reset through auth_flows never sets password_changed_at, so it signs nobody out


ujeenet:
`confirm_password_reset_pool` / `_into` rotate the password hash and nothing else. `password_changed_at` — the column the session middleware uses to invalidate sessions issued before a rotation — is never written, so **a password reset leaves every existing session valid for its full lifetime.**

Password reset is what a user does when they believe their account is compromised. This is the one flow where "sign me out everywhere" is the point, and it is the flow that does not do it.

## The mechanism is live, and the other path uses it

`tenancy/admin.rs:826` rejects a session issued before the last rotation:

```rust
if let Some(ts) = user.password_changed_at {
    if payload.iat < ts.timestamp() {
        return SessionCheck::Anonymous;
    }
}
```

The admin change-password path stamps it (`tenancy/admin.rs:1394`):

```rust
user.password_hash = new_hash;
user.password_changed_at = Some(chrono::Utc::now());
```

The `auth_flows` reset path does not. Its whole write is:

```sql
UPDATE {t} SET {pw} = {p1} WHERE {pk} = {p2}
```

So the same account reached through two documented paths gets two different security outcomes, and the weaker one is the path `docs/auth-flows.md` walks you through.

## Verified, not inferred

A throwaway test against sqlite with a `password_changed_at` column present, through `confirm_password_reset_pool_into` with the framework defaults:

```
PROBE hash rotated: true
PROBE password_changed_at after reset: None
```

The hash rotates; the column stays `NULL`. `NULL` is specifically the value the middleware reads as "never rotated, do not enforce", so it is not merely stale — it disables the check.

## Why this is not a one-line fix

`confirm_password_reset_pool_into` takes `user_table` / `pk_column` / `password_column` and is explicitly for foreign schemas, which may have no such column. Writing it unconditionally would break those callers. Three shapes, none free:

1. **Only the defaults form stamps it.** `confirm_password_reset_pool` knows its table is `rustango_users` and that the column exists. Contained and safe — but `docs/auth-flows.md` walks the reader through `_into` with `"rustango_users", "id", "password_hash"` spelled out, so the documented example would still be the unfixed path. That is most of the value gone.
2. **An optional `password_changed_at_column` parameter.** Explicit and works for custom schemas; another argument on a function that already takes seven, and a silent no-op for anyone who does not pass it.
3. **A `UserModel`-shaped descriptor** carrying the column names together. The right long-term shape, too big for a patch release.

[...]

Issue #1440, opened 2026-09-14
59 more live suites report green against an unreachable database — #1434 fixed 8 of 67


ujeenet:
#1434 fixed the eight `django6_*` suites that swallowed a Postgres connection error. The same idiom is in **59 other test files, covering 203 test functions**, and they behave identically: set `DATABASE_URL` to something unreachable and they report success.

## The line

```rust
let pg = sqlx::PgPool::connect(&url).await.ok()?;   // ← error discarded, caller returns silently
```

`.ok()` on `connect` collapses *"the database is broken"* into *"no database configured"*. The suite then skips and libtest counts it passed.

## Reproduced by execution — two files picked at random from the 59

```
$ DATABASE_URL='postgres://rustango:rustango@127.0.0.1:59999/rustango_test' \
    cargo test -p rustango --test select_related_live --test window_functions_live

select_related_live    test result: ok. 3 passed; 0 failed  ... finished in 90.01s
window_functions_live  test result: ok. 4 passed; 0 failed  ... finished in 120.01s
```

Nothing listens on 59999. Seven tests passed having connected to nothing, and spent 210s doing it.

## Where the split actually is — not where I first thought

Reading `env::var("DATABASE_URL")` is **not** the discriminator. Almost every live suite skips on an absent variable, which is correct and should stay.

The divergence is one call later, at `connect`:

| Behaviour when `DATABASE_URL` is set but the server is unreachable | Files | Sites |
|---|---:|---:|
| `connect(...).await.ok()` → swallowed, suite reports **pass** | **59** | 59 |
| `connect(...).await.unwrap()` / `.expect(...)` → **fails loudly** | 25 | 27 |

So the codebase already does the right thing in 25 files. This is one suite-wide decision answered two ways, and only one of the two can detect a broken database.

### A wrong turn worth recording, so nobody repeats it

Twelve files contain *both* idioms, which looks like self-contradiction and is not. The shape is:

```rust
async fn pool() -> Option<PgPool> {
    let url = std::env::var("DATABASE_URL").ok()?;        // skip if unconfigured — correct
    Some(PgPool::connect(&url).await.unwrap())            // fail if configured-but-broken — correct
}
...
let Some(pool) = pool().await else { return };            // the skip
let url = std::env::var("DATABASE_URL").unwrap();         // safe: presence already established
```

[...]

Issue #1437, opened 2026-09-14
The S3 storage live suite has never run in CI: RUSTANGO_S3_TEST_* is set nowhere, and 25 tests report green instead of saying so


ujeenet:
Five test files gate on `RUSTANGO_S3_TEST_*`. Those variables appear **nowhere in `.github/`** — not as `env:`, not as `${{ secrets.* }}` — and no workflow declares an S3-compatible service. The tests are not `#[ignore]`d, so they run and report as passing.

## Scope — what is and isn't covered

This is **not** "S3 is untested". The signing logic is covered well, and that is the hard part:

`crates/rustango/src/storage/s3.rs:452` — 25 unit tests, including
`signing_key_matches_aws_docs_test_vector`, `sha256_empty_string_matches_known_value`,
`format_amz_date_matches_iso8601_basic`, the `encode_key_*` pair, and four `host_*` cases.
SigV4 is checked against AWS's published vectors.

What has never executed is the **live round-trip** — a real PUT/GET against an S3-compatible
server, presigned-URL fetch, and media upload→serve:

| File | Tests | `#[ignore]`d |
|---|---:|---|
| `media_collections_tags_live.rs` | 16 | 0 |
| `media_live.rs` | 6 | 0 |
| `s3_live_presign.rs` | 3 | 0 |

`media_live_seed.rs` and `s3_live_seed.rs` also read these vars but document `--ignored` and are
deliberately manual — not part of this.

## Reproduced by execution

```
$ cargo test -p rustango --test s3_live_presign
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

Three passed in 0.00s with nothing configured. None of the three carries `#[ignore]`. They are
inside `cargo test --workspace --all-features` (`ci.yml:123`), so CI runs them and counts them
green on every build.

## Why this reads as a gap rather than a decision

CI already spins up every other dependency a live suite needs — Postgres, `mysql:8.0`,
Redis (`redis_live`, `ci.yml:527`), and PostGIS for the geometry suite. S3 is the only one where
the tests exist and the service does not. The test files' own docs even name the intended server:

```
RUSTANGO_S3_TEST_ENDPOINT=http://127.0.0.1:9100      # MinIO
```

So the suite was written to be runnable and the job to run it was never added.

## Secondary: the skip is inconsistent within a single file

`media_collections_tags_live.rs:82-83` says it skipped:

```rust
[...]

PR #1451, opened 2026-09-14
release/v0.57.3 — six issues: logout, password reset, rate limiting, backoff, lockfiles, S3 CI

ujeenet:
Six issues from the documentation-drift sweep. Each fix was verified adversarially — the fix reverted, the matching guard watched to fail — rather than by watching a new test pass.

## Security

- **#1402 — logging out did nothing.** Three defects composed: `JwtBackend` never read the revocation list, never checked `typ` (so a refresh token authenticated as a bearer, with days of life instead of minutes), and `/logout` never touched the refresh token. A user clicked log out, got `204`, and the session survived the full seven-day refresh TTL. Also repairs a regression from #1397 that had stopped `JwtBackend` accepting `JwtLifecycle`'s tokens entirely.
- **#1399 — password reset skipped the password policy and was replayable.** It checked `len() < 8` while `strength_score` — the documented policy — rejects `12345678`; and the link stayed valid for its full TTL *after* the password changed. New `confirm_password_reset_single_use`.
- **#1398 — per-IP rate limiting did nothing behind a proxy.** Every client shared the proxy's bucket. The issue's proposed fix would have been worse than the bug (see below).

## Fixed

- **#1410** — retry backoff was `2s, 4s, 8s` where everything documented `1s, 2s, 4s`. Also: `MAX_ATTEMPTS` is a total-attempt ceiling, not a retry ceiling.
- **#1422** — all eight example lockfiles were stale, not just the one reported.
- **#1437** — the S3 live suites had never executed and reported green on every build.

## Two things worth a reviewer's attention

**#1398 did not land the way the issue proposed.** The issue says to key the limiter on `RealIp`, on the premise that `RealIpLayer` "already has a strategy for this". It does not — `HeaderStrategy` picks which header to read, not whom to believe, and there is no trusted-proxy check anywhere. Keying on `RealIp` would trade a coarse limit for **no limit**: any client can vary `X-Forwarded-For` and mint a fresh bucket per request. Instead `RealIpLayer::trust_proxies([...])` names the hops, and only then does a `TrustedRealIp` appear for the limiter to use. Nothing changes until proxies are declared.

**#1437 uncovered a framework bug, filed as #1450.** The 22 media tests fail against a real Postgres: `SqlValue::Null` binds as text, so a NULL `uploaded_by_id` will not go into a `bigint` column. Media upload is broken on Postgres. Those tests are `#[ignore]`d with the reason recorded — wiring them in would turn CI red for a bug that predates this branch. 52 other `SqlValue::Null` sites share the expression.

## Also filed

- **#1449** — a password reset never writes `password_changed_at`, so it signs nobody out. Distinct from #1338, which is about the comparison rather than the value never being written.

## Deferred to release/v0.57.4

#1440, #1396, #1431, #1409, #1403, plus the two filed above.

## Note on versioning

The workspace is still at `0.57.1` and `[Unreleased]` holds both this branch's entries and 0.57.2's — 0.57.2 merged to develop without a version bump or a cut section. The bump and the section cut are still to do, as a separate step.


ujeenet:
Reviewed the #1398 and #1410 fixes against the tree. Both hold. One follow-up site and two things I checked and cleared.

### A fifth site on #1410

`crates/rustango/tests/jobs_pg_live.rs:138` still reads:

```rust
// First attempt fails immediately, then a 2-second backoff before
// retry. Wait at least one full backoff, then poll for both the
```

`retry_backoff_ms(0)` is now `1_000`, so that describes the removed behaviour. **Stale prose only, not a broken test** — the sleep is 3500ms, comfortably over one full 1s backoff, so it passes and keeps passing. This file isn't in the PR's diff, so it's a follow-up rather than a review blocker.

Worth noting where it sits: the new unit test's own doc comment says *"Described in four places and asserted in none, which is why it survived."* There were five — and the fifth is in a test file, which a docs-scoped sweep doesn't look at.

### Checked and cleared

- **`webhook_delivery.rs:504`** also says "2s backoff", but that subsystem has its own schedule and never calls `retry_backoff_ms`. Same string, different claim — deliberately not swept in.
- **All four locales** carry both halves of #1410, the sequence *and* the total-attempts semantics (`docs/fr/jobs.md:84,258` as "tentatives totales"). My first pass flagged French as missing it; I'd guessed the phrasing. My grep, not a gap.

### On #1398

The fix is safer than the issue that prompted it, and that's worth recording. #1398 proposed keying the limiter on the `RealIp` extension. That would have traded a coarse limit for **no limit**: `X-Forwarded-For` is client-controlled, so any client could mint a fresh bucket per request by varying a header.

`TrustedRealIp` — distinct extension, inserted only when the claim arrives over a declared-trusted hop, defaulting to trusting nobody — is the correct shape, and `rate_limit.rs:99` reads only that. The five tests cover both failure directions, including `a_direct_client_cannot_spoof_its_way_out_of_the_limit` and `without_trust_proxies_the_header_does_not_key_the_limiter`. That's what makes it a guarantee rather than a fix.

ujeenet:
Superseded by #1453, which is the same tree on a clean branch — one commit, no unwanted co-author trailers. This branch's history carried 23 of them, and a squash-merge aggregates those onto `develop` (it already did once, on 6a415053).

The replacement's tree is byte-identical to `release/v0.57.4`'s, verified by comparing `git write-tree` rather than by inspection. Nothing is lost; the branch stays for reference.

Issue #1431, opened 2026-09-14
sql::__macro_internals says "do not import" and has 14 importers, two of them in the flagship example's runtime code


ujeenet:
Found working E2-19 with `rustango-33`. The prose half (the cookbook naming a `fetch_aggregate` that doesn't exist) is a separate doc fix; this is the API question underneath it.

## What the module says

[`sql/mod.rs:84-91`](crates/rustango/src/sql/mod.rs#L84):

> …exist purely to support the macro's `Self::save_on` / `Self::delete_on` / … codegen that needs a generic executor for tenancy schema-mode `SET search_path` scoping. **Not part of the public API; do not import.**

```rust
#[cfg(feature = "postgres")]
#[doc(hidden)]
pub mod __macro_internals {
```

## Who imports it

Fourteen sites, none of them macro expansions:

```
crates/rustango/tests/aggregate_filter_live.rs        fetch_aggregate_on
crates/rustango/tests/pg_aggregates_live.rs           fetch_aggregate_on
crates/rustango/tests/window_functions_live.rs        fetch_aggregate_on
crates/rustango/tests/having_auto_routing_live.rs     fetch_aggregate_on
crates/rustango/tests/groupby_inference_live.rs       fetch_aggregate_on
crates/rustango/tests/prefetch_non_i64_pk_live.rs     fetch_with_prefetch
crates/rustango/tests/prefetch_related_live.rs        fetch_with_prefetch
crates/rustango/tests/where_expr_live.rs              select_rows_on
crates/rustango/tests/order_by_annotate_live.rs       annotate_count_children{,_on}
crates/rustango/examples/cookbook_blog/tests/cookbook_chapter03_orm.rs   fetch_aggregate_on, bulk_insert_on
crates/rustango/examples/cookbook_blog/src/apps/blog/urls.rs:122,168     insert_on, update_on
```

The last two are the ones that matter most: they are **application source in the flagship example**, not tests. A reader copying that pattern is copying an import the crate forbids, and `cargo doc` will not show them the function because the module is `#[doc(hidden)]`.

## Why this is a public-surface question, not a tidy-up

The ten in-tree tests are not misuse. They reach for `fetch_aggregate_on` and `fetch_with_prefetch` because **there is no public way to do what they do** — run an aggregate or a prefetch against a specific executor rather than a pool. `_on` variants are the established name for the executor-taking form; they are simply not exported.

So one of two things is true, and it needs deciding rather than patching:

[...]

Issue #1409, opened 2026-09-13
In-flight jobs die on every SIGTERM: no graceful shutdown on Cli::run, and nothing in the crate handles SIGTERM


ujeenet:
`docs/jobs.md` documents draining in-flight jobs on shutdown. Under any orchestrator the step never runs, because nothing in the crate handles `SIGTERM` and the main server path installs no graceful shutdown at all.

Three separate defects, and fixing any one alone leaves the documented flow broken.

## 1. `Cli::run` installs no graceful shutdown

All three serve calls in [`manage.rs`](crates/rustango/src/manage.rs) — lines 845, 898 and 944 — are:

```rust
axum::serve(listener, app.into_make_service_with_connect_info::<SocketAddr>())
    .await?;
```

No `.with_graceful_shutdown(..)`. The signal terminates the process, so **nothing after `run()` executes**. `docs/jobs.md:162`:

```rust
// 4. Boot the server. Cli::run() blocks until Ctrl-C / SIGTERM.
…
// 5. On shutdown: drain in-flight jobs, then stop.
queue.shutdown().await;
```

Step 5 is unreachable. Not "runs late" or "sometimes misses" — the process is gone before that line.

## 2. The tenancy path has graceful shutdown but ignores `SIGTERM`

[`tenancy/server.rs:197`](crates/rustango/src/tenancy/server.rs#L197) does it properly:

```rust
axum::serve(listener, app)
    .with_graceful_shutdown(shutdown_signal())
```

and then:

```rust
async fn shutdown_signal() {
    let _ = tokio::signal::ctrl_c().await;
}
```

`tokio::signal::ctrl_c()` is **SIGINT only** on Unix. It never resolves for `SIGTERM`, whose default disposition kills the process immediately.

So the two paths are wrong in different ways, and a fix that only adds shutdown to `Cli::run` leaves the tenancy path broken under exactly the orchestrator that motivated the change.

## 3. Nothing in the crate handles `SIGTERM`

```
rg 'signal::unix|SignalKind::terminate' crates/rustango/src/
(no matches)
```

Yet `docs/jobs.md` claims `SIGTERM` twice — at `:162` and again at `:226`:

```rust
tokio::signal::ctrl_c().await?;   // block until Ctrl-C / SIGTERM
```

That comment is a factual error about tokio's API, and it is the snippet a reader copies into their own `main`.

## What this costs

[...]

Issue #1403, opened 2026-09-13
Bulk create is not transactional: a constraint violation mid-batch leaves earlier rows committed behind a 400


ujeenet:
`docs/viewsets.md:509`:

> **Bulk create** is free: `POST` a JSON *array* and every element is inserted in order, validated atomically (one bad element rejects the whole batch).

Validation is atomic. The writes are not, and the sentence reads as though both are.

## What happens

[`viewset/mod.rs:2336`](crates/rustango/src/viewset/mod.rs#L2336):

```rust
for (i, (columns, values)) in prepared.into_iter().enumerate() {
    match insert_and_fetch_one(state, acq, columns, values, pk_field, &fields).await {
        Ok(obj) => created.push(obj),
        Err((code, msg)) => {
            return json_error(code, &format!("bulk entry {i}: {msg}"));
        }
    }
}
```

One `INSERT` per element, no enclosing transaction, and an early return on the first database-level failure. `grep -c 'begin\|transaction\|commit\|rollback'` over that block returns 0.

## The failure

`POST` ten elements. Element five violates a unique constraint or a foreign key — something validation cannot see because it is a property of the database, not of the payload.

- elements 0 through 4 are **committed**
- the caller gets `400 bulk entry 5: …`
- the response body lists no created rows

So the client is told the batch failed, five rows exist, and the response contains nothing identifying them. A naive retry duplicates the first five if the constraint allows it, or fails on element 0 if it does not.

This is worse than a plain non-atomic write, because the error message says "entry 5" while the doc says one bad element rejects the whole batch. Both readings point away from "go and check what actually landed".

## Why validation being atomic is not enough

The doc's claim is true for the checks the serializer can run: shape, types, field validators, cross-field rules. Those are all decided before any write.

Constraint violations are exactly the class it cannot decide up front — uniqueness against rows this request has not inserted yet, foreign keys against rows another transaction may have deleted. That is the class that reaches the loop, and it is the class the loop handles worst.

## Proposed fix

**Wrap the loop in a transaction.** The insert path already takes an acquired handle; it needs a `BEGIN` around the loop and a commit after, with the early return becoming a rollback.

The source comment above the loop explains why this codepath does N round-trips rather than using `bulk_insert_pool` — the dynamic-JSON shape has no typed model handle. That reasoning is sound and unaffected: N round-trips inside one transaction is still N round-trips, and atomicity is the part that was missing.

[...]

Issue #1396, opened 2026-09-13
RUSTANGO_SESSION_SECRET is interpreted three different ways, and check --deploy validates none of them


ujeenet:
`RUSTANGO_SESSION_SECRET` is read in three places, and each interprets it differently. The combination lets a deployment pass `manage check --deploy`, sign JWTs with a below-floor key, and lose every session on restart — all at once, with the only warning on stderr.

## The three readings

| where | decodes base64? | floor applied to |
|---|---|---|
| [`session.rs:82`](crates/rustango/src/session.rs#L82) `from_env_or_random` | **yes** | decoded bytes (≥32), else random per-process key |
| [`auth_routes.rs:93`](crates/rustango/src/tenancy/auth_routes.rs#L93) `build_jwt` | **no** — `.into_bytes()` | raw string bytes (≥32, asserts) |
| [`migrate/manage.rs:4260`](crates/rustango/src/migrate/manage.rs#L4260) `check --deploy` | **no** | raw string length (≥32) |

One variable, three answers to "is this secret long enough".

## A value that breaks all three at once

A 32-character base64 secret. Not exotic — it is what you get from several key generators, and it looks right.

```
32 base64 characters  →  decodes to 24 bytes
```

What each layer does with it:

- **`check --deploy`** — `32 >= 32`, reports **"RUSTANGO_SESSION_SECRET length OK"**. Green.
- **`build_jwt`** — `32 >= 32`, signs access and refresh tokens with the **32 raw characters**. The assert exists specifically to stop a short HMAC key, and it is measuring the encoded form, so a 24-byte secret sails past the 32-byte floor it is enforcing.
- **`session.rs`** — decodes to 24 bytes, `24 < 32`, **falls back to a random per-process key**.

Result: sessions silently stop surviving restarts, JWTs are signed with a key below the floor the code believes it enforces, and the tool whose whole job is catching this says it is fine.

## A non-base64 value is worse

A passphrase or hex string of 40 characters:

- `check --deploy` — length OK
- `build_jwt` — signs with those 40 raw bytes
- `session.rs` — decode fails, **random key**

So JWTs use the configured secret and sessions use a key that changes every restart. Two auth surfaces, one variable, different keys, no error.

## On "silently"

Worth being precise, because the audit note that led here overstated it. `session.rs` does warn — `tracing::warn!` **and** a yellow `eprintln!` saying sessions will not survive a restart. It is not silent at the process.

[...]

PR #1452, opened 2026-09-14
release/v0.57.4 — the things the tests were not testing

ujeenet:
Stacked on #1451, which is still open — GitHub retargets this to `develop` when that merges. Seven issues, each verified adversarially: the fix reverted and the matching guard watched to fail, rather than a new test watched to pass.

## The through-line

#1437 turned on 22 media tests that had never executed. **All 22 failed on first contact with a real database** — that was #1450, and it meant media upload was broken outright on PostgreSQL. A CI-hygiene issue turned out to be hiding a live user-facing break.

That reframed #1440: 66 suites and 204 test functions reporting green against a database that is not there are not a tidiness problem, they are an unknown quantity of real breakage behind a green wall. Expect redness when they first run against a real service, and treat it as the finding.

## Fixed

| | |
|---|---|
| **#1450** | `SqlValue::Null` bound as text, so writing NULL into any non-text column failed on Postgres. **Media upload was broken.** 52 sites shared the expression. Fixed with an OID-0 unspecified NULL — one struct, no enum change, no API break. |
| **#1440** | 66 live suites swallowed a broken database. Unset still skips; set-but-unreachable now panics. A guard recomputes the set from source — it found six files my own grep missed. |
| **#1449** | A password reset never wrote `password_changed_at`, so it signed nobody out — in the one flow whose entire purpose is signing out everywhere. |
| **#1396** | `RUSTANGO_SESSION_SECRET` was read three ways. A 32-character base64 secret is 24 bytes: `check --deploy` reported OK, JWTs were signed below the floor their own assert enforced, and cookies fell back to a random per-process key. |
| **#1431** | `sql::__macro_internals` said "do not import" and had 14 importers, two of them in the flagship example's request handlers. Four of the twelve re-exports were emitted by the macro **zero** times. |
| **#1409** | Nothing handled SIGTERM, so the documented job drain never ran under any orchestrator. New `Cli::on_shutdown` + `shutdown::shutdown_signal()`. |
| **#1403** | Bulk create validated atomically and wrote non-atomically: a constraint violation on element 5 committed 0–4 behind a 400 that named none of them. |

## Two corrections to my own work, both found in review

**#1409 shipped half-wired.** I verified the hook was *called* at five sites without verifying it was *reachable* at five — two sat behind a `Builder::serve` with no graceful shutdown, so the hook ran on a line the process never got to. Fixed in 52f586c9, with a guard that fails the build on any bare `axum::serve`.

**#1396 carries an undocumented breaking change**, now documented: a non-base64 secret of ≥32 raw characters used to sign JWTs perfectly well and now **panics at startup**. A JWT-only API on such a secret was not broken before and will not boot now. Failing closed is right; going unmentioned was not.

## Breaking / behaviour changes

- `Cli::run` returns on signal instead of never returning.
- `check --deploy` and `jwt_router` reject secrets they used to accept (above).
- Live suites now panic rather than skip when their URL is set but unreachable.

## Verification

Tri-dialect where it matters. #1403 ran against **live PostgreSQL 16, MySQL 8 and SQLite** — and my first revert did not reproduce the bug on Postgres, because a constraint violation inside a PG transaction aborts it, so the commit behaved as a rollback. That is not the "before"; before, there was no transaction at all. The faithful revert fails all three with two rows left behind.

## Still open

- **#1338** — the same-second `iat < ts` edge for operators. Distinct from #1449: that one is about the comparison, this was about the value never being written.
- **#1293** — the `_on` family exported by #1431 is PostgreSQL-only. Stated loudly rather than hidden.
- **#1255** — `InMemoryJobQueue::shutdown` aborts rather than drains. #1409 makes it get called; that one makes it do the right thing when it is.

## Note on versioning

The workspace is still at `0.57.1` and `[Unreleased]` holds 0.57.2's, 0.57.3's and this branch's entries. The bump and the section cut are a separate step.


ujeenet:
Superseded by #1453, which is the same tree on a clean branch — one commit, no unwanted co-author trailers. This branch's history carried 23 of them, and a squash-merge aggregates those onto `develop` (it already did once, on 6a415053).

The replacement's tree is byte-identical to `release/v0.57.4`'s, verified by comparing `git write-tree` rather than by inspection. Nothing is lost; the branch stays for reference.

Issue #1461, opened 2026-09-15
Test suite is organized by dialect, not by feature: 30 duplicated stems, and 167 SQLite-only files that a shared harness would make tri-dialect


ujeenet:
The integration suite is 553 files / ~114k LOC, and it is shaped by
dialect rather than by feature. 30 feature stems have a near-duplicate
file per backend (~15,280 LOC), and far more features are tested on
exactly one backend because writing the second and third copy by hand is
expensive.

The duplication is worth removing. The bigger prize is that the same
change turns ~100 SQLite-only ORM tests into tri-dialect ones for
roughly free.

The pattern that does this is already in the repo — `django6_*`, 8 files —
and so is the piece that makes it cheap, `testkit::create_tables_for`. Neither
was ever generalized.

## The shape of the suite

| | files |
| --- | ---: |
| `*_sqlite_live.rs` | 192 |
| `*mysql*` | 16 |
| `*pg*` / `postgres` | 14 |

**167 of the 192 SQLite files have no MySQL or PG counterpart** — 28,244
LOC. 102 of those (17,518 LOC) exercise the ORM query/write path
directly. The largest buckets:

| bucket | files | LOC |
| --- | ---: | ---: |
| `model_*` | 36 | 3,684 |
| `queryset_*` | 26 | 2,575 |
| `viewset_*` | 15 | 3,085 |
| `admin_*` | 20 | 2,740 |
| `operator_*` | 7 | 3,697 |

`model_*` and `queryset_*` together are 62 files of pure ORM-surface
tests that only ever run on SQLite. That is precisely the surface where
MySQL diverges — clause ordering, `TEXT` in an index, `DECIMAL` vs
`INTEGER` aggregate output types.

## The duplication that already exists

30 stems with a per-dialect variant. The clearest triples:

```
mcp_user_keys     [mysql 137] [pg 126] [sqlite 452]
migrate_reconcile [mysql 153] [pg 145] [sqlite 299]
jobs              [mysql 196] [pg 211] [sqlite 175]
bulk_upsert       [mysql 174] [pg 172] [sqlite 193]
explain_pool      [mysql 127] [pg 145] [sqlite 110]
funcs_batch1      [mysql 121] [pg 274] [sqlite 106]
```

I read the `explain_pool` triple end to end. All three share the same
model, the same `select_query()`, and the same three test bodies with the
same assertions. They differ in pool construction, the DDL string, and
which plan-format marker they grep for. Everything else is copied.

[...]

PR #1476, opened 2026-09-15
fix(migrate): MySQL cannot parse the DROP CONSTRAINT this emitted — slice 2 of #1461

ujeenet:
Slice 2 of #1461 — the first real bug the consolidation work turned up.

### The bug

`migrate/diff.rs` special-cased SQLite on `DropCheckConstraint` and `DropCompositeFk` and let everything else fall through to the PostgreSQL shape. Against MySQL 8.0.46:

```
ALTER TABLE `t` DROP CONSTRAINT IF EXISTS `ck_x`
  ERROR 1064 (42000): ... right syntax to use near 'IF EXISTS `ck_x`'
```

MySQL accepts `DROP CONSTRAINT` from 8.0.19 but takes no `IF EXISTS` on any drop-constraint form. **Every MySQL migration that dropped a check constraint or a composite foreign key died with a syntax error.**

The correct branch already existed one module away. `ddl::drop_constraints_sql_with_dialect` does exactly this for per-field FKs, and its comment even spells out the reason: "Postgres `DROP CONSTRAINT`, which additionally accepts `IF EXISTS`". `diff.rs` did not use it.

### Why it survived

Two unit tests asserted the unparseable string and held it in place. Both were named `..._uses_backticks`, and both did check the quoting — the statement around the quoting was simply never run against a server.

That is the distinction this epic keeps running into, and it is worth stating as a rule:

| dialect branch | emission test is… |
|---|---|
| **rejects** the operation (SQLite has no `DROP CONSTRAINT`) | the whole contract — nothing to execute |
| **rewrites** into different SQL (`DROP CHECK`, `DROP FOREIGN KEY`) | proof only that the writer emitted what its author intended |

Every latent bug found so far is a rewrite with no live execution. This was one.

The two MySQL migration suites that do run live could not have caught it either: every migration they build carries only `SchemaChange::CreateTable`, so they never reach this branch. (That corrects the expectation stated in #1474.)

### What is here

1. **The fix** — MySQL gets `DROP CHECK` and `DROP FOREIGN KEY`. Postgres keeps the idempotent `IF EXISTS` form. One behavioural difference is left standing because the engine leaves it standing: the Postgres drop is idempotent, the MySQL one is not (error 3821 when absent). Noted at both call sites.
2. **The two unit tests corrected and renamed**, with the 1064 in their doc comments so nobody "simplifies" them back.
3. **`migrate_drop_constraint_tri`** — the live suite that did not exist. Adds a CHECK, proves it bites, drops it, proves the row is accepted again; then the same round trip for a composite FK. It asserts on **enforcement**, not on rendered SQL.

### Acceptance

Against MySQL 8.0.46 and PostgreSQL 16 with both URLs set — reverting each branch in turn:

| reverted | `tri_mysql` | `tri_postgres` | `tri_sqlite` |
|---|---|---|---|
| `DropCheckConstraint` | **FAILS**, `ERROR 1064 ... near 'IF EXISTS \`ck_dc_tri_n_positive\`'` | ok | ok |
| `DropCompositeFk` | **FAILS**, `ERROR 1064 ... near 'IF EXISTS \`fk_dc_tri_child_parent\`'` | ok | ok |
| nothing | ok | ok | ok |

PostgreSQL and SQLite staying green is the point — the bug was MySQL-only, and a guard that went red everywhere would be testing something else.

### Two things writing it surfaced

**`fresh_table` does not create composite-unique indexes.** `emit_tables` creates the table, its FK constraints and its comments; `testkit/mod.rs:140` states outright that composite-unique indexes are not among them. So a `unique_together` model built by the harness carries no composite uniqueness and nothing can reference it — PostgreSQL answers "no unique constraint matching given keys", MySQL answers 1822. The setup here renders the index through `CreateIndex` rather than hand-writing it. **The harness gap is left for its own change** — worth flagging, because any converted suite relying on `unique_together` is currently asserting against a constraint that does not exist.

**The guard from #1474 was too narrow.** It watched `*_mysql_live.rs` only, but `*_tri.rs` suites need a `mysql_live` line just as much — and three had already shipped without one: `values_tri`, `regex_tri`, `testkit_matrix_forms_tri`.

A missing line is *worse* for a `_tri` suite. A `*_mysql_live` suite with no line does nothing and reports `0 passed`, which is at least odd-looking. A `*_tri` suite keeps running its PostgreSQL and SQLite arms in `postgres_test` and reports a healthy pass count, while the one arm it was converted to gain never runs.

Widened to both families, renamed `every_mysql_arm_runs_in_ci`, all four missing lines added. The three pre-existing suites pass on MySQL.

### Verification

```
every_mysql_arm_runs_in_ci     5 passed   (fails naming all 4 gaps before the ci.yml lines)
migrate_drop_constraint_tri    6 passed   (postgres + mysql + sqlite)
values_tri                    15 passed
regex_tri                      9 passed
testkit_matrix_forms_tri       6 passed
migrate lib + migrate_diff + migrate_invert   222 + 37 + 22 passed
```

Workflow YAML re-parsed: 21 jobs, all 20 non-gate jobs still carry `needs: gate`.


ujeenet:
Superseded by a PR against `release/v0.57.6` — this work belongs on the release branch, not `develop`. The branch also carried an incomplete commit (a `git add` named a pre-rename path and staged nothing, so the ci.yml lines and the guard widening were missing); that is fixed in the replacement.

Issue #1412, opened 2026-09-13
An unresolvable list_display name is dropped silently, and a fully broken list_display looks unconfigured


ujeenet:
A name in `list_display` that resolves to nothing is dropped without a word. The column is simply absent from the admin, and nothing says why.

## Verified

[`admin/views.rs:457`](crates/rustango/src/admin/views.rs#L457), the code's own comment:

```rust
// Names that match none are silently dropped. Empty
// `list_display` falls back to every scalar field (today's
// behavior).
```

Resolution is a `filter_map` over four kinds of thing: declared fields, `register_admin_computed!` entries, `generic_fk` declarations, and dotted JSON paths. Anything matching none of them falls out of the iterator.

`docs/admin.md:125` documents the key without mentioning it:

> Columns shown in the list, in order. FK columns render the target's `display` value. Computed columns (see below) can be named here. Empty = every scalar field.

## The failure

`list_display = "id, titel, status"` renders two columns. No error, no warning, no log line. The admin looks like it works.

Three ways this bites in practice, all quiet:

- **A typo.** The column is missing and the cause is invisible; the natural conclusion is that the field is not supported in `list_display`.
- **A renamed model field.** The admin silently loses the column at the same commit that renames it — and nothing in the test suite fails, because the admin still renders.
- **A computed field whose `register_admin_computed!` did not run** — a `#[cfg]`-gated module, a registration behind a feature. Same silence, harder cause.

The empty-means-everything fallback makes it worse at the extreme: if every name fails to resolve, the list is empty and the code falls back to showing **every scalar field**. So a fully-broken `list_display` looks like a deliberately unconfigured one.

## Fix

Two options, and I would do both.

1. **Warn at resolution.** `tracing::warn!` naming the unresolved names and the model, once per render or once per process. Cheap, no behaviour change, and it turns an invisible failure into a greppable one.
2. **Fail loudly where it is cheap to.** The admin config is declared, not user input, so an unresolvable name is a programming error rather than a runtime condition. If there is a natural startup or registration point, refusing there is better than warning per request — it turns a silently wrong admin into a failure at the moment someone made the mistake.

If neither is wanted, `docs/admin.md:125` must at least say that unrecognised names are dropped, so the behaviour is discoverable from the page documenting the key.

## Note on scope

[...]

PR #1488, opened 2026-09-16
docs: Tier B + C findings that still held against 0.57.6

ujeenet:
From the `.claude/docs-audit` Tier C set ("stated as the inverse of reality" — the costliest class, because a confident wrong statement does not prompt anyone to check).

Re-verified all twelve open findings against 0.57.6 rather than taking the audit at face value. **Four still held.**

| | claim | reality |
|---|---|---|
| **C-14** | `testing.md`: the test helpers are "always compiled" | `test_client` is `#[cfg(feature = "admin")]`. `test_assertions` and `test_db` genuinely are ungated — right about two of three, which is how it read as safe |
| **C-13** | `auth-backends.md`: same claim for `rustango::auth_backends` | needs the internal `_async_trait` feature, pulled in by `cache`/`email`/`jobs`/`oauth2` |
| **C-25** | `sso.md` intro: the member flow "may opt into" auto-provisioning | `MemberAuthConfig::default()` sets `auto_provision: true`. The page says so correctly and in bold at `:153` — only the intro sentence, the one a reader meets first, was soft |
| **C-26** | `sso.md`: the flow cookie is `Secure` "on HTTPS" | nothing inspects the request scheme |

C-26 is the one worth reading twice. `secure_cookies()` takes an explicit `security.secure_cookies` policy — default `true` on the `manage` path, so it fails closed — and otherwise falls back to "secure on the prod tier" from `RUSTANGO_ENV`. An HTTPS deployment that sets neither ships the cookie **without** `Secure`, and an operator who read this sentence would never think to check.

### Already fixed — no change needed

- **C-01** (the pool/executor rule) now carries the corrected wording, the reads-vs-writes table and the #1293 reference the audit asked for.
- **C-02** now states outright that the `…Layer` types are config structs, *not* `tower::Layer`, and that `.layer(..)` does not compile on them.

### Still open in Tier C, not addressed here

C-08, C-11, C-18, C-20, C-22, C-23, C-24 — each needs its own verification against the current tree, and I would rather do that properly than in bulk.

Docs guards green: `docs_links` 5, `docs_versions` 14, `docs_live_suite_counts` 2.


ujeenet:
Extended with **Tier B** (broken examples). Re-verified the eight open findings; three still held:

- **B-06** — `with_rollback`'s example used `|tx| async move { … }`, but the bound is `for<'tx> FnOnce(&'tx mut PoolTx<'_>) -> Pin<Box<dyn Future<…> + Send + 'tx>>`. The `Box::pin` is what lets the closure borrow `tx` across its own awaits, so the snippet could not compile. The module's own rustdoc had it right; the page now matches.
- **B-09** — the i18n page built `let translator = Translator::new(…)` then called `register(&mut tera, translator.clone())`. `Translator` holds three `RwLock`s and is not `Clone`; `register` takes `Arc<Translator>`. "Cheap to clone-share" is true of the `Arc`, not the value.
- **B-18** — the `tenant_user_extension` README exported `DATABASE_URL` *after* running `init-tenancy`. Only `help`, `version`, `docs`, `startapp`, `showurls`, `showmodels` are pool-free, so it failed with `missing env var` before doing anything. The step is now gone entirely, because `init-tenancy` is a no-op kept for old scripts — its own `--help` says so.

Already fixed, no change needed: **B-05** (uses the `RouterExt` form now) and **B-12** (`.cookie_name()` shows `.to_string()`).

Also worth recording: **Tier A is complete.** All 21 verified against 0.57.6 — A-13 through A-19 and A-21 are all fixed (async cache ctor, transactional bulk create via #1403, dump banner on stderr, `Cli::on_shutdown`, 1s backoff with a test that says "not 2s", `list_display` warns via #1412, `init-tenancy` help corrected, `match_nothing` is public and the doc uses it). A-20 was a wrong premise. Only **A-06** held → #1486.

PR #1584, opened 2026-09-18
ci: six test suites compiled to empty crates where anyone could see

### LINKED ISSUE #1572: CI: every test suite 0.57.7 adds compiles out of the only ungated job
From the crew review of the 0.57.7 release (run `pr-1544`): tests-002, -003, -004, -005, -009. Measured, with controls.

This is #1437 and #1548's family, one layer up: not a suite that skips silently, but a suite that **is not compiled at all** where anyone would notice.

## The gap

`default = ["postgres", "batteries"]`, and `batteries` pulls in neither `sqlite`, `testkit` nor `mysql`. Every suite this release adds or rewrites is gated on one of those three:

`media_router_requires_authorization` · `media_soft_delete_consistency` · `media_perms_default_policy` · `fk_on_delete_live` · `media_tags_mysql_live` · `assert_num_queries_sqlite_live`

`clippy` is the only ungated job that compiles `tests/**`, and it runs on default features — so all six compile to **empty crates** there. They execute only in `postgres_test` / `sqlite_live` / `mysql_live`, every one `needs: gate`.

**Consequence:** a test file that does not build merges green on an unlabelled PR into `develop`. That is verbatim the regression `ci.yml:246`'s own comment says `--all-targets` was added to stop — and `--all-targets` cannot help when the file is `#![cfg]`'d out.

The new `feature_combos` media rows do not cover it either: `cargo check` does not build test targets (measured), and the `--lib` entries are the library.

## Guards satisfied by their own comments

Two structural guards match text that a **comment** also satisfies, so commenting a step out leaves them green:

| mutation | guard | result |
|---|---|---|
| comment out the two `s3_live` media `--test` steps | `the_media_live_suites_are_named_in_the_s3_job` | **survives** |
| delete the `sqlite,media*` matrix rows, keep the comment | `feature_combos_covers_the_media_axis` | **survives** |
| *delete* either outright | both | die correctly |

So the guards check that the text exists somewhere in the file, not that the job runs the step.

## Also

`tests-009`: the new `pull_request: branches:` filter means a PR into a feature branch runs **nothing** — including `fmt`, `clippy`, `guards`. That is the intended saving, but it puts the structural guards out of reach of exactly the stacked work that tends to touch CI wiring. Worth a deliberate decision rather than a side effect.

## Fix sketch

Add an ungated job that does `cargo check --tests` (or `--no-run`) with a feature set that actually includes `sqlite,testkit` — cheap, since it never executes anything — and make the two structural guards match a step that is *not* commented out (anchor on the line start, or parse the YAML).

## Verified working, do not re-check

`every_mysql_arm_runs_in_ci` genuinely covers the new MySQL suite, the `s3_live` env block and both live-job URLs — confirmed by commenting the `media_tags_mysql_live` step out and watching it fail.

ujeenet:
Closes #1572. First item of the 0.58.0 slice, taken first because it is the guard that protects everything after it.

## The hole

`default = ["postgres", "batteries"]`, and `batteries` pulls in neither `sqlite`, `testkit` nor `mysql`. A suite gated on one of those compiles to an **empty crate** without it — it builds, it reports `ok`, and it contains nothing.

`clippy` was the only ungated job that compiled `tests/**` at all, and it runs `--features tenancy` on top of the defaults. So on an unlabelled PR into `develop`, these six were neither run nor built:

`media_router_requires_authorization` · `media_soft_delete_consistency` · `media_perms_default_policy` · `fk_on_delete_live` · `media_tags_mysql_live` · `assert_num_queries_sqlite_live`

**Every guard the 0.57.7 security pass added is in that list.** A change breaking one of their signatures merged with a green tick and first failed post-merge, on develop's push run. That is #1437 one layer down: not a suite that *skips* silently, but a suite that is never *built* where anyone would look.

## Demonstrated, with a control

The same deliberately broken line appended to `assert_num_queries_sqlite_live.rs`:

| command | result |
|---|---|
| `cargo test -p rustango --features tenancy --no-run` (what `clippy` builds) | **0 errors** — the suite is `cfg`'d out and invisible |
| `cargo test --workspace --all-features --no-run` (the new job) | **fails**, `E0425` |

The control is the point. Without the first row this is a job that passes, which proves nothing.

## The guard is about gating, not about the command

The way this rots is not someone deleting the job. It is someone adding `needs: gate` to it — it is the slowest ungated job, that will look like an easy saving, and the moment it is gated it stops running on exactly the pull requests it exists for.

So `every_test_suite_compiles_in_ci` asserts that *some* job builds the full test surface **and** that the job doing it has no `needs:`. Both mutations kill it:

| mutation | fails on |
|---|---|
| add `needs: gate` to `tests_compile` | the `needs:` assertion — names `["tests_compile", "postgres_test"]` as all-gated |
| delete `tests_compile` entirely | the same one, since `postgres_test` is the only remaining builder and it is gated |

Unlike `every_mysql_arm_runs_in_ci` there is no list of suites to fall behind the directory: `--workspace --all-features` is by construction every suite under every feature.

## Cost, stated

`--all-features` shares no build cache with `clippy` or `guards` (both `--features tenancy`), so this adds roughly one cold build per PR. That is the price of the six suites being compiled anywhere an unlabelled PR can see, and it is already what `.githooks/pre-push` asks of a developer locally.

## Verified

`every_test_suite_compiles_in_ci` 1/1 · `cargo fmt --all --check` clean · both mutations confirmed failing before restore.


ujeenet:
## Review round applied — the guard was fail-open in five ways

A seven-angle crew review (`.claude/review/runs/pr-1584/report.md`, 22 findings) found the **job** correct and the **guard** wrong. Fixed in `3379f994`.

The root cause is the one this PR exists to prevent, one level up: I mutation-tested the guard against the two changes I had in mind — `needs:` and deleting the job — rather than against the property it claims to establish.

### What got past it, all reproduced by execution

| mutation | was | now |
|---|---|---|
| job-level `if:` (how `gate` itself is gated) | ok | **FAILED** |
| `continue-on-error: true` (as `windows_test` uses) | ok | **FAILED** |
| append `--lib` | ok | **FAILED** |
| append `--exclude rustango` | ok | **FAILED** |
| append `--test docs_links` | ok | **FAILED** |
| delete job + trailing `#` comment elsewhere | ok | **FAILED** |
| `needs: gate` | FAILED | FAILED |

It also failed **closed** on two correct workflows — the backslash-continuation style this file uses for multi-flag cargo invocations, and `cargo check --workspace --all-features --tests`. A guard that cries wolf gets deleted, which ends where one that misses ends. Both now pass, along with a trailing comment on the job header.

### Three false claims in the comment, corrected

- **"six suites"** was the count the 0.57.7 media pass happened to touch. The real figure is **275 of 559**.
- **`testkit` was named as a gap.** It is not — the self dev-dependency turns it on for any `cargo test` build. The real unmentioned gaps are `mcp` (16 suites), `cache-redis`, `cache-page`, `passkey`, `email-smtp`, `admin-sso`.
- **"already what `.githooks/pre-push` asks of a developer locally"** — the entire justification offered for the extra build, and wrong. The hook runs `cargo check` (lib+bins; `--tests` is opt-in) and `cargo test --lib`. Neither compiles `tests/**`, so it would not have caught #1572 either. That sentence was also in the commit message and this PR's description.

### The job did not fully close #1572

Six suites are gated on a **negated** feature, so `--all-features` compiles them out by construction. A second command now covers the five `not(postgres)` ones — the same command `sqlite_litmus` runs, gated. `oauth2_router_standalone` needs a third shape and remains only in `feature_combos`; that gap is stated rather than hidden behind a doc claiming completeness.

### Also

`timeout-minutes: 45` — the job is reachable from a fork PR and is the first ungated one to build three `build.rs` scripts (`libsqlite3-sys` with `bundled` compiles the SQLite C amalgamation); the default is six hours. And the `gate` and `guards` headers both enumerate the ungated jobs and were both stale.

### Verified

Seven breaking mutations fail, three correct shapes pass, `on:`-block leakage into the job parse closed (23 jobs parsed, none spurious), full `guards` job green locally — 51 tests across nine suites — `cargo fmt --all --check` clean.

**Still unverified:** `performance` never ran, so nobody measured the build time or cache behaviour. Finding 4 above shows the stated justification for that cost was false, which makes the gap more material, not less.

Issue #1598, opened 2026-09-18
migrate: the #1588 DropTable guard reached one of three loops — CHECK and EXCLUDE drops still reproduce it, and suppression broke rollback


ujeenet:
Two defects in the merged #1589 fix, found by a crew review of #1591 (`correctness-001`, `correctness-002`). Both verified against the code; neither is in an open PR.

## 1. The guard reached one loop of three

#1589 stopped `detect_changes` emitting `DropIndex` for an index whose table the same migration drops. `detect_changes` has **three** structurally identical loops, and the other two were not touched:

| loop | line | guarded? |
|---|---|---|
| dropped indexes | `diff.rs:336` | ✅ added by #1589 |
| dropped CHECK constraints | `diff.rs:421` | ❌ |
| dropped EXCLUDE constraints | `diff.rs:450` | ❌ |
| dropped composite FKs | `diff.rs:499` | ✅ already correct, and documents why |

So dropping a model that declares a table-level `CHECK` emits `DropTable` (`:310`) and then `DropCheckConstraint` (`:421`) — and reproduces #1588 exactly:

- **MySQL** — `DROP TABLE` auto-commits, `ALTER TABLE t DROP CHECK ck` fails **1146**, the ledger row is never written, the re-run fails **1051**. The unrecoverable state #1588 exists to prevent.
- **PostgreSQL** — `42P01` inside the atomic transaction. Clean rollback, but the model can never be dropped at all.
- **SQLite** — the renderer refuses outright.

**Fix:** the same one-line condition the index loop got, in both places — `&& current.table(&c.table).is_some()`. `DropCompositeFk` at `:499` already carries it, so the omission is visible in the file.

## 2. Suppression traded a loud failure for a silent one

With `DropIndex` suppressed, a drop-model migration's forward list is `[DropTable]`. `invert` yields `[CreateTable]` — and the `CreateTable` render arm (`diff.rs:806`) emits the table plus FK constraints and **no index DDL**. Verified.

So rolling back now **succeeds** and restores the table without any of its indexes, UNIQUE ones included, while the predecessor snapshot still lists them. `makemigrations` sees no drift, and nothing ever recreates them.

Before #1589 the same rollback failed loudly, because `CreateIndex` was ordered before `CreateTable`.

[...]

Issue #1588, opened 2026-09-18
MySQL migrations: DropIndex is unrenderable (and makemigrations emits it), plus partial-DDL failures leave schema and ledger out of sync


ujeenet:
## Summary

`#559` is closed, but two of the gaps it tracked are still live and both were
hit in one afternoon on a MySQL tenant. The error messages emitted to users
still point at `#559`, so anyone who follows the breadcrumb lands on a closed
issue.

---

## 1. `DropIndex` is unrenderable on MySQL

`SchemaChange::DropIndex` (`crates/rustango/src/migrate/diff.rs:140`) carries
only `name`. MySQL needs `DROP INDEX <name> ON <table>`, so the renderer
fails loudly rather than emitting anything
(`crates/rustango/src/migrate/diff.rs:1017-1029`):

```
DropIndex for `rpt_goal_dimension_idx` is not yet supported on dialect
`mysql`. MySQL requires the table name (`DROP INDEX <name> ON <table>`) but
the `SchemaChange::DropIndex` variant only carries `name`. ... Tracked in #559.
```

### Why it's easy to hit

`makemigrations` **generates** these ops itself. Dropping a model emits
`DropTable` plus one `DropIndex` per index — so a perfectly ordinary
"remove a table" migration is un-appliable on MySQL straight out of the
generator:

```json
"forward": [
  {"schema": {"DropTable": "rpt_goal"}},
  {"schema": {"DropIndex": {"name": "rpt_goal_dimension_idx"}}},
  {"schema": {"DropIndex": {"name": "rpt_goal_fiscal_year_idx"}}},
  {"schema": {"DropIndex": {"name": "rpt_goal_unit_idx"}}}
]
```

Those three `DropIndex` ops are also **redundant** — MySQL (and Postgres)
drop a table's indexes with the table.

### Suggested fixes, smallest first

1. Don't emit `DropIndex` for indexes belonging to a table dropped in the
   same migration. Fixes the common case with no schema change.
2. Promote the variant to `DropIndex { name, table }` as the error text
   already anticipates, and render MySQL's form.

---

## 2. A failed MySQL migration leaves partial DDL applied *and* no ledger row

`runner.rs:2204` already warns that MySQL auto-commits DDL. What that means
in practice is worse than the warning suggests:

The migration above ran `DropTable` **successfully**, then failed on the
first `DropIndex`. Result:

- the table was **gone**
- the migration was recorded as **failed**, so no ledger row
[...]

PR #1601, opened 2026-09-18
release: 0.57.8 — promote develop → main

ujeenet:
**Merge with _Rebase and merge_** — `required_linear_history` on the `main` ruleset rejects a merge commit. `develop` is 17 commits ahead and holds no content `main` lacks (checked line by line; the only differences the other way are older version strings and pre-correction copies of comments this release updated).

## 0.57.8 — migrations, and the guards that should have caught them

**An ordinary "remove a model" migration was unapplyable on MySQL** (#1588), and when it failed it left the schema and the ledger permanently disagreeing — reported from a live tenant, reproduced twice. `DropTable` auto-commits, the following `DropIndex` fails 1146, the ledger row is never written, and the re-run fails 1051 with nothing able to reconcile it.

**And the fix for that was wrong twice over** (#1598) — stated plainly in the changelog rather than smoothed over:

- it suppressed the dependent drop in one of `detect_changes`' **three** identical loops, so a model with a table-level `CHECK` still reproduced the original failure;
- and suppression broke rollback: `[DropTable]` inverts to `[CreateTable]`, which renders no index DDL, so rolling back *succeeded* and silently dropped every index while the snapshot still listed them.

Both fixed by **ordering rather than suppression**, guarded by a round-trip test — the property, not the op list.

**CI integrity.** An audit of what this repo actually guards found the machinery meant to catch this class was not running: six structural guards on no pull request at all, 275 of 559 suites compiling to empty crates in the only ungated job that built `tests/**`, two example crates compiled by nothing, and a guard whose own check was a frozen literal. The `guards` job went from **9 targets to 16**.

**Media.** `media::router` is now documented as the internal management API, and `MediaManager::public_url(id)` is the supported way to show an uploaded image on a public page.

## ⚠️ Breaking

`MigrateError` is `#[non_exhaustive]` and gained `PartiallyApplied`. Only affects exhaustive matches — add a `_ =>` arm. `UPGRADING.md` has the note.

Shipped as a micro rather than 0.58.0 to match this repo's practice (0.57.7 shipped a runtime-breaking change as a patch).

## Verified

96 checks green on the release PR — the `release/**` branch name triggers the full push matrix, so this had the complete run, not the ungated subset. `bin/bump-version.sh` handled 6 manifest sites, four doc languages and 11 regenerated lockfiles.


ujeenet:
Superseded by the single-commit promote — see below. `develop` and `main` can no longer converge by rebase: the 0.57.7 promote was rebase-merged, so main carries rewritten copies of develop's commits under different SHAs (main's `cdd91f62` and develop's `d3debb1f` have identical trees). Replaying develop's seventeen commits therefore tries to re-apply work main already has, which is what this PR was conflicting on.

PR #1602, opened 2026-09-18
release: 0.57.8 — promote develop → main

ujeenet:
Replaces #1601, which conflicted. **Merge with _Rebase and merge_** — one commit onto `main`, so it is a clean fast-forward.

## Why one commit instead of develop's seventeen

`develop` and `main` can no longer converge by rebase. The 0.57.7 promote was rebase-merged, which rewrote develop's commits onto main under new SHAs — main's `cdd91f62` and develop's `d3debb1f` have **identical trees and different identities**. So replaying develop's commits now tries to re-apply work main already has, and conflicts on it. That is what #1601 hit.

**This will recur every release** while `required_linear_history` forbids a merge commit and `allowed_merge_methods` forbids a squash. Worth settling deliberately — see below.

This commit's tree is `origin/develop` **exactly**, verified byte-for-byte rather than merged: `git diff promote-0578 origin/develop` is empty.

## 0.57.8

**An ordinary "remove a model" migration was unapplyable on MySQL** (#1588), and when it failed it left the schema and the ledger permanently disagreeing — reported from a live tenant, reproduced twice.

**And the fix for that was wrong twice over** (#1598): it patched one of `detect_changes`' three identical loops, so a model with a table-level `CHECK` still reproduced the failure; and its suppression approach broke rollback, silently restoring a table without its indexes where the old behaviour had failed loudly. Both fixed by ordering rather than suppression, guarded by a round-trip test.

**CI integrity.** An audit found six structural guards running on no pull request, 275 of 559 suites compiling to empty crates, two example crates compiled by nothing, and a guard whose own check was a frozen literal. The `guards` job went from 9 targets to 16.

**Media.** `media::router` is documented as the internal management API; `MediaManager::public_url(id)` is the supported way to show an uploaded image publicly.

## ⚠️ Breaking

`MigrateError` is `#[non_exhaustive]` and gained `PartiallyApplied` — affects exhaustive matches only. `UPGRADING.md` has the note.

## The recurring problem, for a decision

Three options, none free:

1. **Keep doing this** — a single tree-carrying commit per promote. Works, and main's history becomes one commit per release rather than the full development log.
2. **Allow `merge` on `main`** by dropping `required_linear_history` there. Restores the pre-0.57.7 behaviour, where main carried develop's commits plus a merge commit.
3. **Allow `squash`** in `allowed_merge_methods` for `main`, which is effectively option 1 done by the UI.

Option 2 is closest to what the repo did for its first fifty releases. Option 1 is what this PR does and needs no settings change.

## Verified

96 checks green on the release PR (the `release/**` branch name triggers the full push matrix). Tree identical to `develop`.


ujeenet:
Superseded by the proper merge-commit promote — `required_linear_history` is off `main` now, so the single-commit workaround this PR used is no longer needed. A real two-parent merge also re-establishes shared ancestry between the branches, which is what stops the next promote conflicting the same way.

Issue #1586, opened 2026-09-18
CI: a PR into a feature branch runs no workflow at all, so stacked PRs get no signal


ujeenet:
Found by the crew review of #1584 (`correctness-004`), and ruled **out of scope for that PR** because it is pre-existing — it arrived with `d3debb1f`, that PR's own base. Filing it so it is not re-derived by the next review.

## What happens

`.github/workflows/ci.yml`'s `pull_request:` trigger carries a `branches:` filter, and `branches:` filters on the **base** — "where is it going", not "where did it come from". The list is `main`, `develop` and the release globs.

So a pull request whose base is an ordinary feature branch matches nothing and **the workflow does not run at all**. Not "the matrix is deferred" — nothing runs: no `fmt`, no `clippy`, no `guards`, no `doc`, no `deny`, no `deny-examples`, no `lockfiles`, no `trivy`.

The comment above the filter states this outcome plainly and accepts it, on the reasoning that such merges "cannot reach `main` without a second PR that runs them again". That reasoning is sound for the *eventual* state and wrong for the *working* state.

## Why it bites

Stacking is the normal shape here whenever a change is too large for one review — and it is about to become more common, because `required_linear_history` on `develop` now makes assembling work in a release branch impractical, which pushes the same work into stacks based on each other.

Concretely, in a stack `A → B → C` where B's base is A and C's base is B:

- B and C get **no signal at all** while they are open and being reviewed. Not a deferred matrix — no formatting check, no lint, no advisory scan.
- The structural guards are exactly what a stacked PR most needs, because they catch the class of mistake that only shows up in combination: `every_mysql_arm_runs_in_ci`, `docs_contract`, `docs_versions`, and now `every_test_suite_compiles_in_ci`.
- `deny` and `trivy` are security checks. A dependency added in B is unscanned until B lands in A and A is re-opened against `develop`.

The "it runs later" argument also assumes the stack is eventually flattened into a PR against `develop`. When B merges into A, the run on A covers the *combined* diff — so the first time anyone sees B's own changes checked is after they are already merged into A. That is the same post-merge-discovery pattern #1437, #1548 and #1572 are all instances of.

## Options

[...]

Issue #1542, opened 2026-09-17
signed_url fails open on a clock error — every expired URL verifies


ujeenet:
**Severity:** Low · **Area:** `signed_url` · **Found on:** `develop` @ `ca726393`

`signed_url::current_unix_secs` fails **open**. [signed_url.rs:143-147](crates/rustango/src/signed_url.rs#L143-L147):

```rust
fn current_unix_secs() -> u64 {
    SystemTime::now()
        .duration_since(UNIX_EPOCH)
        .map_or(0, |t| t.as_secs())
}
```

`duration_since` errors when the system clock is before the Unix epoch. On that path `now_secs` is `0`, and the expiry check in `verify_at` is `if now_secs > exp` ([signed_url.rs:117-122](crates/rustango/src/signed_url.rs#L117-L122)) — `0 > exp` is false for every `exp`, so **every expired URL verifies as valid**.

## Severity

Low, honestly: a pre-1970 clock needs a misconfigured container, a dead RTC on bare metal, or a deliberately skewed host. It is not a remote attack. It is filed because the failure mode is the wrong one — a security primitive's error path should deny, and this one grants. The same shape in a deployment where clock sync is the thing that broke is exactly when you least want signatures to stop expiring.

The rest of the module is solid and worth not disturbing: the signature covers the `expires` param so it cannot be stripped, `canonicalize` sorts by key **and** value so param reordering cannot forge, and the comparison is `subtle::ConstantTimeEq`.

## Suggested fix

Return `Result`, or saturate the other way:

```rust
fn current_unix_secs() -> u64 {
    SystemTime::now().duration_since(UNIX_EPOCH).map_or(u64::MAX, |t| t.as_secs())
}
```

`u64::MAX` makes a broken clock expire everything, which is the safe direction. A `Result` that surfaces `SignedUrlError::ClockError` is better still — it tells the operator what broke instead of producing a confusing wave of 403s.

## Verification

`verify_at(url, secret, 0)` against a URL whose `expires` is in the past must return `Expired`. That is a direct unit test — `verify_at` already takes the time.

Note `expires_at` is optional in `sign`, so a URL signed with `None` never expires by design. That is a documented choice and separate from this.

Issue #1540, opened 2026-09-17
CI runs jlumbroso/free-disk-space@main — a third-party action on a mutable ref


ujeenet:
**Severity:** Low · **Area:** CI · **Found on:** `develop` @ `ca726393`

[.github/workflows/ci.yml:318](.github/workflows/ci.yml#L318) runs a third-party action pinned to a **mutable branch ref**:

```yaml
- name: Free disk space
  uses: jlumbroso/free-disk-space@main
```

Whoever controls that repository can change what `main` points at, at any time, and it executes on the runner on the next CI run — before `actions/checkout`, so it is the first thing in the job. Every other `uses:` in the workflow is at least tag-pinned (`actions/checkout@v4`, `dtolnay/rust-toolchain@1.88`, `Swatinem/rust-cache@v2`, `EmbarkStudios/cargo-deny-action@v2`).

## Blast radius — genuinely limited

The workflow sets `permissions: contents: read` at the top level, with a comment explaining exactly why ([ci.yml:4-9](.github/workflows/ci.yml#L4-L9)), so the `GITHUB_TOKEN` cannot write to the repo. This is already the right call and it is what keeps this at Low rather than High.

What is still reachable: the runner environment, the job's `DATABASE_URL` (a local service container, low value), and — the one that matters — the `Swatinem/rust-cache` cache, which this step runs before and could poison for subsequent runs on the same key. A poisoned build cache is a real persistence foothold even under a read-only token.

## Suggested fix

Pin to a full commit SHA with the version in a trailing comment, the standard shape:

```yaml
- uses: jlumbroso/free-disk-space@<40-char-sha>  # v1.3.1
```

Worth doing for the other third-party actions at the same time (`Swatinem/rust-cache`, `EmbarkStudios/cargo-deny-action`, `dtolnay/rust-toolchain`); `actions/*` is first-party and lower risk but SHA-pinning it costs nothing. Dependabot keeps SHA pins updated, which is also on the list in #1492.

## Verification

`grep -rn 'uses:' .github/workflows/ | grep -v '@[0-9a-f]\{40\}'` returns nothing. Worth adding as a CI step so the next added action cannot skip it — same shape as the workflow-coverage guard added in #1474.

Issue #1539, opened 2026-09-17
MCP raw-key cache keeps a revoked credential working for up to 60 seconds


ujeenet:
**Severity:** Low · **Area:** `mcp::auth` · **Found on:** `develop` @ `ca726393`

A successful raw-API-key verification is cached for 60 seconds with no invalidation hook, so a revoked agent credential keeps authenticating for up to a minute after `revoke_user_key_pool` / `rotate_agent_secret_pool` returns success.

[mcp/auth.rs:411-452](crates/rustango/src/mcp/auth.rs#L411-L452):

```rust
const RAW_KEY_CACHE_TTL: std::time::Duration = std::time::Duration::from_secs(60);
const RAW_KEY_CACHE_CAP: usize = 256;

fn raw_key_cache_get(key: &[u8; 32]) -> Option<(i64, Option<i64>)> {
    let cache = raw_key_cache().lock().ok()?;
    let (identity, verified_at) = cache.get(key)?;
    (verified_at.elapsed() < RAW_KEY_CACHE_TTL).then_some(*identity)
}
```

The cache itself is well built — the key is `SHA-256(tenant \0 token)` so it is tenant-scoped ([:424-433](crates/rustango/src/mcp/auth.rs#L424-L433)), it is bounded at 256 with an expiry sweep before a hard clear, and the argon2-skip it buys is a real win on a hot MCP transport. The only thing missing is the revocation edge.

## Impact

Bounded to 60 seconds and one process, so this is not urgent. It matters because revocation is the operator's emergency lever: during an incident, "the key is revoked" should be true when the command returns, not a minute later, and not "on whichever pods happen to have it cached". The API returning success while the credential still works is the part worth fixing.

## Suggested fix

1. `revoke_user_key_pool` and `rotate_agent_secret_pool` call an invalidation that drops matching entries. The cache is keyed by a hash of the raw token, which the revoke path does not have — so either key on `(tenant, agent_id)` as a secondary index, or expose `raw_key_cache_clear()` and call it on any revoke (cheap: worst case is 256 extra argon2 verifications).
2. Document the window on the const, whichever way it lands.
3. Consider dropping the TTL to ~10s; the argon2-skip benefit is mostly in the first few seconds of a burst.

## Verification

Authenticate, revoke, authenticate again — the second call must fail. That test fails on `develop` today.

PR #1604, opened 2026-09-19
0.57.9 — thirteen small holes

### LINKED ISSUE #1526: url_has_allowed_host_and_scheme accepts /\evil.example — open redirect the sibling safe_next already rejects
**Severity:** High · **Area:** `urls` · **Found on:** `develop` @ `ca726393`

`urls::url_has_allowed_host_and_scheme` accepts `/\evil.example/x` as a safe same-origin URL. Chrome, Firefox and Safari normalise `\` to `/` in the path of a special-scheme URL, so a `Location: /\evil.example/x` sends the browser to `https://evil.example/x`. Open redirect.

## The gap

[urls.rs:262-269](crates/rustango/src/urls.rs#L262-L269):

```rust
// Rule 3 — protocol-relative / Windows-path injection.
if trimmed.starts_with("//") || trimmed.starts_with("\\\\") || trimmed.starts_with('\\') {
    return false;
}
// Rule 4 — relative URLs.
if trimmed.starts_with('/') {
    return true;
}
```

Rule 3 catches a leading `\`, a leading `\\`, and a leading `//`. It does not catch `/` **followed by** `\`. Rule 4 then returns `true`.

Django, which this function is a documented parity port of, handles it by checking the input twice:

```python
return _url_has_allowed_host_and_scheme(url, allowed_hosts, require_https) and \
       _url_has_allowed_host_and_scheme(url.replace("\\", "/"), allowed_hosts, require_https)
```

with the comment *"Chrome treats \ completely as / in paths"*.

## Why it survived

The unit tests cover `\\evil.com\x` and `\backslash\path` — both leading-backslash forms, both caught by Rule 3 ([urls.rs:1208-1215](crates/rustango/src/urls.rs#L1208-L1215)). There is no test for the `/\` shape.

The sibling helper `auth_decorators::safe_next` gets this right — `is_safe_path` explicitly rejects `starts_with("/\\")` ([auth_decorators.rs:742-752](crates/rustango/src/auth_decorators.rs#L742-L752)) and its doc calls out *"Backslash variants the browser normalizes to a host (`/\evil.example/x`)"*. So the framework knows about the shape; the Django-parity public helper is the one that misses it — and that is the one an integrator coming from Django reaches for by name.

## Suggested fix

Run the whole rule set over both `trimmed` and `trimmed.replace('\\', "/")`, Django-style. Also worth folding in: Rule 4 should reject `/\` directly so the two helpers agree.

## Verification

Add to the existing test module:

```rust
assert!(!url_has_allowed_host_and_scheme("/\\evil.com/x", &[], false));
assert!(!url_has_allowed_host_and_scheme("/\\/evil.com/x", &[], false));
assert!(!url_has_allowed_host_and_scheme("/a/\\evil.com", &["example.com"], false));
```

Confirm it fails on `develop` before applying the fix, not just that it passes after.

ujeenet:
Thirteen verified defects from the 2026-09-18 triage, each small enough that the fix is
smaller than the argument for it. One commit per concern.

Every item was re-verified in the source before being worked; where the issue disagreed
with the code, the code won, and the commit says so.

## Security

| # | What was wrong |
|---|---|
| #1526 | `url_has_allowed_host_and_scheme` accepted `/\evil.com/x`. Browsers rewrite `\`→`/`, so it left as protocol-relative and the redirect landed off-site. Now validates both the raw and rewritten forms, as Django does. |
| #1525 | Six paths put raw sqlx text — table, constraint, column list, database host — into unauthenticated 5xx bodies. One `server_error_body`: log the cause, publish it only on the debug tier. |
| #1537 | `brand_logo_url` / `brand_favicon_url` went into `src=`/`href=` through `\| safe` on seven templates, including both login pages and the admin sidebar. `brand_css` keeps its `\| safe`, and the test records why. |
| #1542 | Signed-URL expiry failed **open**: an unreadable clock read as `0`, making `now > exp` false for every timestamp. Now `u64::MAX`. |
| #1539 | The MCP raw-key cache skipped argon2 for 60s and nothing cleared it, so a revoked key kept working for a minute after the command reported success. |
| #1529 | The CSRF same-origin check compared only the host and fell back to the raw header, so `Origin: example.com`, `Origin: null` and `http://` against an https site all passed. *(Partial — the default-off behaviour is the breaking half and stays open.)* |
| #1536 | Doc drift only: the module claimed OsRng throughout while `get_random_string` uses `thread_rng`. Both are CSPRNGs; no weak token was ever minted. |

## Correctness

| # | What was wrong |
|---|---|
| #1465 | `#[rustango::main]` installed a subscriber before `main`, so `Cli::with_logging()` always lost and every scaffolded project silently discarded `[logging]`. New `logging = false` opt-out, and `Setup::install` now reports a lost race instead of swallowing it. |
| #1307 | Every fresh non-tenancy project got a registry-scoped `0001` that `migrate` never applies and that later collides with `relation already exists`. The registry snapshot is never empty — the shared tables are pulled into every scope by name. |
| #1220 | `AlterColumnMaxLength` refused SQLite, where `VARCHAR(n)` and `TEXT` share an affinity and the change is a no-op. MySQL still errors, asserted as a control. |
| #1401 | The ViewSet has served RFC 10008 QUERY since #1112 and the spec never said so, so no generated client could reach it. |

## CI

| # | What was wrong |
|---|---|
| #1586 | A `branches:` filter on `pull_request` meant a PR into a feature branch ran **no jobs at all**. Every stacked PR merged with zero signal. |
| #1540 | `jlumbroso/free-disk-space@main` — a mutable ref on an action that runs before checkout and can write the rust-cache. Pinned to v2.0.0's SHA. The issue reported one call site; there were two. |

## Verification

Every fix was revert-tested: the change was undone, the guard was watched to **fail**, and
only then restored. Where a first guard passed against the reverted code it was rewritten —
that happened three times and each commit records it:

- #1542's guard drove `verify_at` with an explicit time, so the fallback it was meant to
  cover never ran. Split into `clock_secs` and asserted directly.
- #1307's guard tested the predicate, not the caller, and passed against the naive
  snapshot-based form. Moved to `make_migrations_system` itself.
- #1537 and #1401 grew controls so an over-broad fix could not pass for the wrong reason.

Ran locally: the full 18-target `guards` job (which caught an undocumented tracing target
that would otherwise have gone red in CI), `cargo test --workspace --all-features --no-run`,
and the lib suites under `sqlite`, `sqlite,tenancy`, `sqlite,template_views`, `mcp` and
`csrf,tenancy`.

**Not run locally:** anything needing live Postgres, MySQL or MinIO. Nothing here is DDL
against a live database — #1220 and #1307 are rendering and generation decisions, covered by
unit and file-level guards.

Two new guards are registered in the `guards` job; `every_structural_guard_runs_in_ci`
failed on both until they were, which is that check working.

Closes #1526, #1525, #1537, #1542, #1539, #1536, #1465, #1307, #1220, #1401, #1586, #1540
Refs #1529


ujeenet:
## Review fixes — the three inert fixes are now real

Crew review (`.claude/review/runs/pr-1604/report.md`, 55 findings / 7 angles) found that **three of this PR's security fixes did not do what their commits claimed**, plus a merge blocker. All four are fixed in `d8ae00ca`, `2a193019`, `3ceefe48`.

### The merge blocker — this PR's green CI was misleading

`feature_combos` is gated and was **skipped** on this PR. It runs `-D warnings`, and the three new `error.rs` helpers were ungated while every caller sits behind `admin`/`tenancy`:

```
error: constant `OPAQUE_SERVER_ERROR` is never used
error: function `debug_details_enabled` is never used
error: function `server_error_body` is never used
```

Merging would have turned `develop` red. Fixed with `cfg_attr(not(...), allow(dead_code))` — not `#[cfg]`, so widening a caller's gate later can't turn it into a missing-item error. Verified against **all seven matrix rows I can run locally**; my first attempt gated on `_tera` and broke `sqlite,template_views`, which enables it but calls none of this.

### #1525 was inert by default

`server_error_body` asked `debug_details_enabled()`, which answers *"render the dev overlay?"* and returns true whenever `RUSTANGO_ENV` isn't prod. Nothing sets that variable, so **a default deployment still published driver text**. Disclosure is now its own opt-in, `RUSTANGO_DISCLOSE_ERRORS`, defaulting off.

The old test pinned `RUSTANGO_ENV=prod` — it asserted the one configuration where the fix was active. New tests assert with nothing set, and with the overlay tier explicitly on.

### #1526 fixed a helper with zero callers — and there were **three** live holes, not two

The review found two hand-rolled sanitizers still accepting `/\evil.example/x`. A ratchet added with the fix found a **third**: `member_auth::safe_landing`, which never checked `://` at all and feeds `LOCATION` via `redirect_with_cookie`.

All three now delegate to `auth_decorators::safe_next`. Five copies of one security rule with three different behaviours was the actual defect; the ratchet is what stops a sixth.

### #1539 was net-zero, and two of its three call sites never needed it

The hook cleared a process-local cache from the `manage` CLI — a different process from the server. It also evicted **every tenant's** agent N, since ids are per-tenant sequences. And `verify_raw_agent_credential` already re-runs the liveness check on cache hits, so deletion and deactivation were covered; rotation was the only gap, and the only case the hook couldn't reach.

Hook removed. Rotation is now detected from `secret_rotated_at` on the row the liveness query already reads — no extra query, and it works across processes. `user_id` is read fresh rather than cached, which also closes `tenancy-003`.

### On the guards

Mutation testing found four of this PR's new guards didn't fail when their defect was restored. Each fix above was verified by reverting it and watching the *specific* guard fail:

- the default-posture test fails on the old tier, while the old prod test still passes
- each sanitizer's guard fails on its own hand-rolled form, naming the payload
- the new rotation test fails, while `revoked_key_is_refused_immediately` still passes — proving it never was coverage

Ran: all 19 guards, `cargo test --workspace --all-features --no-run`, the 7 feature-combo rows, and lib suites under `sqlite,tenancy`, `sqlite,tenancy,sso`, `sqlite,admin`, `mcp`.

### Still open, deliberately

The remaining findings — `or_500!` and ~19 unconverted 5xx sites, the empty `[Unreleased]`, `migrate --plan` printing PG DDL for the SQLite no-op, the branding guard testing tera rather than the admin registry, and the uncapped `create_many` — are tracked in the report. The branding-guard rebuild and the CHANGELOG are the two I'd do before merge; say the word.

ujeenet:
## Second crew review — the fix delta (`188eeada..f29a46cd`)

**Verdict: not mergeable.** The first review found three security fixes that did
nothing. These four commits replaced them with three that do *something* — and each is
defeated by a mechanism the commit message did not consider. `postgres_test` is also
red, from this delta.

### The structural result that matters most

**No ungated CI job runs `--lib`.** 17 of the 18 assertions this delta adds are lib unit
tests; `guards` passes explicit `--test` names, `tests_compile` is `--no-run`, and
`postgres_test` / `feature_combos` / `windows_test` all `needs: gate`.

So on an unlabelled PR the only new guard that executes is `one_redirect_sanitizer` —
and mutation shows **that passes with the canonical redirect rule inverted**. This PR's
green CI was measuring almost none of its own work. One `--lib` line in the ungated
`guards` job fixes it (`--features tenancy,sso`, no database needed).

### Three fixes, three defeats

| # | What defeats it |
|---|---|
| **#1539** | **Wrong at zero clock skew.** `verified_at` is stamped *after* Argon2 has already read the row, and `agents.rs:175` stamps `secret_rotated_at` *before* awaiting `save_pool`. Both biases push the same way, so a rotation committing inside that ~12 ms window is accepted **and cached fresh for the full 60 s TTL**. Skew is secondary — this fails with perfect clocks. |
| **#1526** | **Consolidated onto the weaker copy.** `is_safe_path` checks three prefixes and nothing else; the copy hardened in #1526 — then declared callerless — rejects control characters anywhere. `?next=%2F%09%2Fevil.example%2Fx` survives, reaches `Location`, and the browser strips the TAB (WHATWG URL §4.1) → `//evil.example/x`. Measured end to end through the real crate. |
| **#1525** | **One 5xx body still discloses.** `template_views.rs:3007` serves `error_page_html` — `Display`, `Debug`, full source chain — on the still-default-ON overlay tier, eleven lines above a comment saying that must not be published. |

Plus **`postgres_test` red**: removing `| safe` made Tera escape `/` to `&#x2F;`, and
`branding_live.rs:304` asserts the raw substring `src="/__brand__/`. The rendered page is
**correct** — the HTML parser decodes entity references in attribute values before URL
resolution — but the test asserts a spelling rather than the property.

### One change closes seven findings

Stop comparing clocks. Compare **`secret_prefix`**, which `agent_auth_state_pool`
already reads, against the prefix already parsed from the presented token at
`auth.rs:351`:

| closes | how |
|---|---|
| rotation (zero-skew ordering) | new credential → new prefix → mismatch |
| both skew directions | no clock involved |
| dialect precision | string equality, precision-independent |
| **cross-tenant redemption** | a foreign tenant's agent has a different prefix |

Three reviewers proposed value-comparison independently — the stamp, `secret_hash`, and
the prefix. The prefix subsumes all three, because it is precision-independent *and*
random per credential.

That last row is its own high: **a warm cache entry is redeemed against whatever pool is
passed.** `authenticate_agent_by_prefix_pool` — the only thing binding a token to an
agent — runs *only on a miss*. Demonstrated with two tenants: purge tenant `acme`
(freeing the slug), re-provision it, and for 60 s a credential from the destroyed tenant
authenticates as agent 1 of the new one. Pre-existing structure, but this delta sharpened
it by sourcing `user_id` from the foreign row.

### Two guards that do not guard

- `a_verification_older_than_the_rotation_is_refused` — the test named for the #1539 fix —
  **asserts `chrono`'s `>` operator** on two locally-built timestamps and contains zero
  calls to `verify_raw_agent_credential`. Deleting the rotation check leaves it and all
  six siblings green.
- `one_redirect_sanitizer`'s canonical-rule check greps for the literal
  `starts_with("/\\")`, which survives inverting the branch it guards. 5 of 7 hand-rolled
  spellings also survive the main pattern.

### Checked rather than assumed

Three claims from the commit messages were verified, not taken: the query count per
authenticated MCP request is **unchanged** (both call paths counted, statements
character-identical); `disclose_server_errors` is **cheaper** than what it replaced
(186→94 ns); the cache value **did not grow** (both tuples exactly 40 B — `DateTime<Utc>`
fits the padding `Option<i64>` occupied).

`secret_rotated_at` also round-trips correctly on **live PG 16, MySQL 8.0 and SQLite** —
both of my briefed suspicions were wrong. MySQL emits `datetime(6)`, and #1464 does not
touch this column. But dialects found something better: **#1464's own proposed fix would
silently reopen #1539** with a one-second window, and nothing connects the two issues.
(Moot once the fix compares `secret_prefix`.)

### One dispute, resolved against the earlier run

The shared env lock. The first run reported *reproducing* a failure from `logging.rs`'s
private mutex. This run could not construct one in 25 full runs at 16 threads plus 20
focused runs, even with the window widened 400 ms. The fix is correct by construction —
three mutexes over one process-global variable serialize nothing — but the evidence was
overstated, and nothing would catch its re-introduction.

### Limits

No live HTTP request against a running server, so the open redirect is proven at the
library boundary plus the header bytes rather than from a browser. Skew magnitudes are
reasoned from the mechanism, not measured. Nobody opened the OpenAPI viewer router.

Issue #1605, opened 2026-09-19
bump-version.sh: the rewrite handles eight claim shapes, the verification checks six


ujeenet:
`bin/bump-version.sh` rewrites version claims with a `perl` pass, then verifies
no stale claim survives. The entry that introduced the verification says the two
handle the same set of shapes, and the script's own comment repeats it. They no
longer do.

## What diverged

The perl pass has **eight** substitutions (`bin/bump-version.sh:145-152`). Six
take the full version (`$o` → `$n`); the last two take the **series** version
(`$os` → `$ns`):

```perl
s/(\brustango\s*=\s*")\Q$os\E(?![0-9.])/$1$ns/g;
s/(\brustango\s*=\s*\{[^\n]*?version\s*=\s*")\Q$os\E(?![0-9.])/$1$ns/g;
```

These landed in `aa7ea7d4` ("keep crate install examples on current series").

The verification's `CLAIM` alternation still has **six** branches, and it greps
for `$OLD` — the full version. A series claim never contains the full version,
so those two substitutions have no verification coverage at all: a stale
`rustango = "0.57"` is rewritten if the pass reaches the file, and nothing
checks that it was.

## Why it matters

The verification exists because the first cut ran a bare `git grep` and exited 1
on prose, so a bump could not complete. The fix was to ask the narrower question
the rewrite asks. That property held when both sides had six shapes; adding two
to one side without the other quietly reintroduces the asymmetry, in the
direction that fails *open* — a missed claim ships rather than failing the bump.

## Not urgent

No release has shipped a stale series claim that I can find, and the pass and
the verification both run on every bump. This is the guard being narrower than
the thing it guards, not a live defect.

## Fix

Either extend `CLAIM` with the two series shapes and grep for `$OLD_SERIES`
alongside `$OLD`, or derive both the substitutions and the alternation from one
list so they cannot drift again. The second is the reason the shapes were
enumerated in the first place.

Found while correcting the changelog's "five claim shapes" to six (#1522); the
script's comment has been updated to describe the gap accurately in the
meantime.

ujeenet:
Fixed in `400068e8` (#1639). The two series shapes are verified now, on a series bump only.

Issue #1543, opened 2026-09-17
Five doc claims on security paths are contradicted by the code beside them


ujeenet:
**Severity:** Low · **Area:** docs / `docs-truth` backlog · **Found on:** `develop` @ `ca726393`

Four doc comments on security- or reliability-relevant code state something the code beside them does not do. Each is the kind of drift the `docs-truth` skill exists for: the claim reads as a guarantee, a reviewer who trusts it stops looking, and the next change is made on a false premise.

Filed together because they are one class and one sweep; each has a home issue where the behavioural fix lives, linked below.

## 1. "no information leak in the response body" — and the body has the error in it

[template_views.rs:3005-3021](crates/rustango/src/template_views.rs#L3005-L3021). The comment introducing the non-debug branch says *"The plain-text path stays the production default — same stderr/tracing breadcrumbs, no information leak in the response body."* The branch is:

```rust
(StatusCode::INTERNAL_SERVER_ERROR, format!("template render error: {e}"))
```

Behavioural fix belongs to the driver-errors-in-500-bodies issue.

## 2. `FileCache` claims an atomicity `std::fs::write` does not provide

[cache/mod.rs:933-936](crates/rustango/src/cache/mod.rs#L933-L936): *"per-process atomicity via `std::fs::write` (atomic per-call on most filesystems)"*. `std::fs::write` is `O_TRUNC` + write; it is not atomic on any common filesystem. The sentence sits directly under a correct "Limitations vs Django" note about Django's lock file, which makes it read as considered rather than assumed.

Behavioural fix belongs to the FileCache issue.

## 3. `crypto.rs` says two private copies were consolidated; both still exist

[crypto.rs:77-81](crates/rustango/src/crypto.rs#L77-L81) claims `totp::constant_time_eq` and `forms::csrf::constant_time_eq` were folded into `crypto::constant_time_compare`. Both are still present and still called ([totp.rs:201](crates/rustango/src/totp.rs#L201), [forms/csrf.rs:645](crates/rustango/src/forms/csrf.rs#L645)).

Behavioural fix belongs to the constant-time issue.

## 4. `Tenant::pool()` says the schema-mode pool is not cached; it has been since #1235

[extractors/tenant.rs:154-161](crates/rustango/src/extractors/tenant.rs#L154-L161):

[...]

Issue #1523, opened 2026-09-17
CHANGELOG 0.57.6: EXPLAIN ANALYZE `actual time=` credited to MySQL 8.0.46; explain_pool_tri says 8.0.18+


ujeenet:
Lowest-severity item in this batch, and arguably readable as an observation rather than
a capability threshold — but this changelog is scrupulous about version floors elsewhere
in the same document, so the inconsistency has a cost.

## What the changelog claims

0.57.6, Testing:

> **`explain_pool_tri` asserted nothing on two of three backends.** … Making it
> two-sided immediately showed the MySQL arm's stated reason was false: it claimed the
> framework does not opt into `EXPLAIN ANALYZE` and that the plan stays an estimate, and
> **MySQL 8.0.46** reports real `actual time=` measurements. The claim is corrected.

## What the code says

`crates/rustango/tests/explain_pool_tri.rs:122` — the corrected `because` that this
entry is describing:

```rust
because "the framework does opt in, and MySQL 8.0.18+ executes the query and \
         reports `actual time=..` per node just as PostgreSQL does. This arm …"
```

8.0.18 is the release in which MySQL shipped `EXPLAIN ANALYZE`. 8.0.46 is the server
the CI leg happened to run against.

## Why it is worth one line

The same document states version floors precisely everywhere else — "MySQL accepts
`DROP CONSTRAINT` from 8.0.19", "Both renames are portable (MySQL 8.0, SQLite 3.25+)"
— so a reader on 8.0.30 has fair cause to read 8.0.46 as the threshold and conclude
their server does not produce `actual time=` output.

## Fix

Cite 8.0.18+, or say "the 8.0.46 server in CI" to mark it as an observation rather than
a floor.

---

Found proofreading the 0.57.5/0.57.6 changelog against the code; audit id **CL-10**.

ujeenet:
Fixed in #1606, shipped in 0.57.10 (`8e147272`, tagged `v0.57.10`).

Closing by hand: #1606 landed as a single squashed commit whose body reads `Closes #1502, #1503, …`, and GitHub only honours the keyword before the first number in a list like that, so only #1502 auto-closed.

Issue #1522, opened 2026-09-17
CHANGELOG 0.57.6: bump-version.sh's verification checks six claim shapes, not five


ujeenet:
Minor, but the entry's whole point is that the rewrite and the verification must handle
the same set of shapes — and the sentence asserting they agree miscounts one of them.

## What the changelog claims

0.57.6, Fixed (the `bin/bump-version.sh` entry):

> … the final verification still ran a bare `git grep` for the old version and exited 1
> on every prose hit … The verification now checks the **five** claim shapes the rewrite
> handles.

## What the script does

Six. The `perl` rewrite has six substitutions, `bin/bump-version.sh:143-148`:

```perl
s/(version\s*=\s*")\Q$o\E(?![0-9.])/$1$n/g;                       # version = "X"
s/("version"\s*:\s*")\Q$o\E(?![0-9.])/$1$n/g;                     # "version": "X"
s/(version:\s+)\Q$o\E(?![0-9.])/$1$n/g;                           # version:   X
s/(--version\s+"?)\Q$o\E(?![0-9.])/$1$n/g;                        # --version X
s/^(rustango\s+)\Q$o\E(?![0-9.])/$1$n/gm;                         # rustango X
s/^((?:cargo-)?rustango[a-z-]*\s*=\s*")\Q$o\E(?![0-9.])/$1$n/gm;  # rustango = "X"
```

The `CLAIM` alternation the verification uses has the same six branches
(`bin/bump-version.sh:225`), and the script says so in both places — the comment above
the rewrite enumerates six shapes, and the comment above the verification reads:

> The alternation below is the same **six** shapes the perl pass rewrites, and nothing
> else.

## Fix

five → six.

---

Found proofreading the 0.57.5/0.57.6 changelog against the code; audit id **CL-09**.

ujeenet:
Fixed in #1606, shipped in 0.57.10 (`8e147272`, tagged `v0.57.10`).

Closing by hand: #1606 landed as a single squashed commit whose body reads `Closes #1502, #1503, …`, and GitHub only honours the keyword before the first number in a list like that, so only #1502 auto-closed.

PR #1606, opened 2026-09-19
0.57.10 — fifteen places the prose disagrees with the tree

### LINKED ISSUE #1502: logging: "pretty" still names the default format in 8 sites, now that Format::Pretty is a different formatter
`0.57.6` made `Format::Pretty` a genuinely distinct formatter. Before it, `pretty` was accepted and rendered as the default, so "pretty" was a loose synonym for "the normal terminal output" and the docs used it that way. Now `Setup::new()` builds `Format::Full`, and `Pretty` is multi-line, one field per line — a different thing.

Every site that used the old loose sense is now a wrong statement, and two of them are the first thing a reader sees.

## The sites

All state, in one form or another, "the default (non-JSON) format is pretty":

| Site | Text | Reality |
|---|---|---|
| `crates/rustango/src/logging.rs:8` | `rustango::logging::setup();  // env-filter, pretty, "info,sqlx=warn"` | `setup()` → `Setup::new().install()` → `Format::Full` |
| `crates/rustango/src/logging.rs:36` | "Install the canonical dev logger: **pretty format**, env-filter from …" | same |
| `crates/rustango/src/logging.rs:246` | "New builder with defaults: **pretty format**, `"info,sqlx=warn"` filter…" | the next four lines are `format: Format::Full` |
| `crates/rustango/src/logging.rs:552` | "JSON in prod, **pretty** in dev" (`setup_for_env`) | `Setup::new()` → `Full`; only the JSON branch is conditional |
| `crates/rustango/src/logging.rs:584,594` | tests named `should_use_pretty_for_other_envs`, `should_use_pretty_when_unset` | they assert `!should_use_json_for_env()`, which is a claim about JSON, not about `Pretty` |
| `docs/logging.md:71` | `rustango::logging::setup();   // pretty, env-filter, "info,sqlx=warn"` | in a snippet readers paste |
| `docs/logging.md:194` | "`setup_for_env()` reads `RUSTANGO_ENV` and picks JSON when it is `prod` or `production`, **pretty** otherwise" | `full` otherwise |

`logging.rs:246` is the sharpest: the doc line and the field initialiser it describes are four lines apart and disagree.

## Why it matters now and did not before

The same release added the table at `docs/logging.md:165-167` that teaches `full` and `pretty` as different output, and a test (`rendered::format_variants_render_differently`) that fails if they ever render the same. So the repo now asserts in one place that these are distinct and says in six others that the default is the one it is not.

A reader who takes `Setup::new()`'s doc at face value and then sets `format = "pretty"` expecting no change gets multi-line output.

## Separately, and not fixed by the same edit

`crates/rustango/src/logging.rs:292` — `without_targets`: "Hide event targets (the module path) in **pretty** output."

`fmt_layer` applies `.with_target(targets)` to `base` *before* the format match, so it affects all four formatters, and the no-file path does the same. This one is wrong about scope, not about the default, so a sweep that only replaces "pretty" with "full" leaves it wrong in a new way.

## Verification

By inspection of `crates/rustango/src/logging.rs` at `v0.57.6` (`Setup::new`, `fmt_layer`, `setup`, `setup_for_env`, `should_use_json_for_env`). No execution needed — the disagreement is between a doc line and the initialiser directly beneath it.

ujeenet:
**Stacked on #1604** — review that first; this PR's own diff is the single `docs:`
commit at the top. Retarget to `develop` once #1604 lands.

Fifteen doc/comment corrections in one commit. No behaviour change, except two test
gates that were wrong (below). Every claim was re-checked against the code, and several
turned out to be wrong differently than filed — those are called out in the commit
message.

## The ones with a real cost

| # | Reading it and acting on it gets you |
|---|---|
| **#1518** | A 403ing admin form. The note tells you to fix a custom `Method::POST` view, then gives a remedy that only works for the *other* half — `chrome_context` is `pub(crate)`, so `csrf_input` is undefined there and Tera renders it empty. Both halves now have their own fix. |
| **#1504** | The belief that your configured `redact_query_params` key is logged in cleartext. The comment states the design that was **rejected** during the fix, with a reason that is false. |
| **#1543** ④ | Steered away from `Tenant::pool()`, the always-tenant-scoped accessor, by a cost warning #1235 made obsolete. |
| **#1517** | A retry predicate that hard-fails mid-deploy. The entry gives 3821 for both MySQL drops; `DROP FOREIGN KEY` raises 1091. |
| **#1272** | `cannot find PgPool in sqlx` — the mount snippet hardcoded it. |
| **#1507** ① | An attempt to "fix" SQLite FK naming so `DROP CONSTRAINT` starts working. It never will; SQLite has no such statement, and the framework already names every FK. |

The rest are counts and omissions: #1502, #1503, #1505, #1506, #1515, #1519, #1520,
#1521, #1522, #1523, and the remaining #1543 / #1507 items.

## Two that are not documentation

- **#1460 gap 4** — `tx_methods_sqlite_live` was gated `sqlite` **and** `postgres`, so
  the only end-to-end proof the transaction path works *without* Postgres never ran in
  a sqlite-only or mysql-only build. Nothing in the file touches Postgres. All three
  tests pass under `sqlite,tenancy`.
- **The `192 SQLite suites` figure had no guard**, unlike its siblings in `matrix.rs`
  which are recomputed from the tree. Corrected to 188 *and* brought under the existing
  recount — verified by putting 192 back and watching it fail.

## Found while working, filed not fixed

**#1605** — `bump-version.sh`'s perl pass now rewrites **eight** claim shapes while the
verification alternation still checks six, so the two series-version substitutions are
unverified. The script's own comment asserted six and was silently false; it now
describes the gap. Fixing it is a behaviour change to the release tooling and does not
belong in a docs PR.

## Verification

Full 18-target `guards` job, the lib suite under `sqlite,tenancy` (3703 tests), and
`cargo doc` to confirm the `Self::install` intra-doc link from #1503 resolves now that
the comment is back on `Setup`.

One correction to my own work: the first draft of the #1272 fix used a `connect_pool`
function that does not exist. `Pool::connect` is the real one, and it is what
`docs/getting-started.md` already shows.

Closes #1502, #1503, #1504, #1505, #1506, #1507, #1515, #1517, #1518, #1519, #1520, #1521, #1522, #1523, #1543

Refs #1272 — the mount and test snippets are swept in all four locales now, but the issue's first half (the scaffolder no longer emitting `admin_router`) is untouched by this PR, so it stays open for that.


ujeenet:
## Crew review — 6 angles, 29 findings (5 high, 13 medium, 14 low)

Reviewed at `6c7bdcdf` against this PR's own base (`fix/0.57.9-small-holes`, merge-base `188eeada`), not `develop`. Angles: correctness, security, tenancy, dialects, tests, conventions. Performance dropped — no runtime code in the diff.

### Verdict

**The substance is sound and the fix list is short, but five corrections overshot.**

The arithmetic is genuinely good. Every recount in this PR was independently re-derived and holds: 188 suites, eleven lockfiles, fourteen call sites across twelve files, six-verified/eight-rewritten claim shapes, the eight-job ungated set at v0.57.6. All eight cited issue numbers exist and say what the prose claims. Both test-gate changes are correct and were verified by execution.

The damage has one shape: **five places where a correction replaced a wrong claim with a new absolute the tree does not support.** Because these arrive labelled as re-verified, they are harder to catch than what they replaced. Three of the five make a security- or cost-relevant guarantee that is false on a real configuration.

**Fix first: #1 below.** It is the only finding where the new prose tells an auditor a secret is redacted when, on a supported config, it is rendered in cleartext.

---

### High

**1. `crates/rustango/src/tracing_layer.rs:17` — span redaction claim is false on the `access_log = false` path**

The new header states unconditionally that `url.query` passes through the access log's *configured* `redact_query_params`. It does on the normal path. But `[logging] access_log = false` → `access_log_layer()` returns `None` (`manage.rs:1043`) → `mount_observability` takes the `None` arm and builds `TracingLayer::new()` with **defaults** (`access_log.rs:245`), while the span stays mounted. Redaction is exact-match, so `[audit] redact_query_params = ["invite_token"]` renders that key in cleartext on every span.

The two settings live in **different config sections**, so turning off the access log silently voids an audit setting. The replaced comment described the rejected design and was wrong — but it was wrong in the safe direction.

*Fix:* qualify the sentence, or better, pass the configured list through on the `None` arm too — which is what the sentence already promises. `tests` notes the harness for this exists and already takes the `None` argument; deleting the `.redact()` call leaves every logging test green.

**2. `crates/rustango/src/extractors/tenant.rs:157` — "Cost: none worth avoiding" is false past the scoped-pool cap**

`max_cached_scoped_pools` defaults to **64**. At the cap, `scoped_pool` logs a warning and returns the freshly built pool **without inserting it** (`pools.rs:994-1003`) — the pre-#1235 behaviour, permanently, for every tenant that never gets a slot. Tenant 65 pays a fresh PG pool build on every request.

The deleted warning was a verbatim description of the above-cap case. Three other places in the crate still state it correctly — the field doc (`pools.rs:132-136`), `scoped_pool`'s own doc (`pools.rs:957`), and the warning string itself — so this puts the crate at four-against-one with itself. The fallback is already pinned by `scoped_pool_past_the_cap_falls_back_instead_of_failing` (`pools_live.rs:547`).

*Fix:* restore the cost note, scoped to the above-cap case.

**3. `crates/rustango/src/request_id.rs:10` — the new header tells readers to drop `req_id=`; its own snippet then logs no id**

Reproduced by execution. `record()` is `Span::current().record("request_id", id)` — a no-op when the current span has no such field. The Quick start directly below the new section mounts `.request_id(...)` and **no `TracingLayer`**, and the PR deleted the explicit `req_id` from it. As written it renders `INFO rid: handling /me` with no id anywhere; adding `.layer(TracingLayer::new())` makes `request_id="…"` appear.

A reader following this loses correlation ids entirely — worse than the "tedious" pattern it replaced. `handle()`'s own source comment and `tracing_layer.rs` both state the caveat correctly; only the page a reader lands on overclaims.

*Fix:* add the layer to the snippet, or scope the claim to "when the tracing layer is mounted".

**4. `docs/{de,es,fr}/viewsets.md:256` — the `PgPool` → `sql::Pool` fix landed in English only, and the PR closes the issue**

All three translations still carry `use rustango::sql::sqlx::PgPool;` and `PgPool::connect(...)` — confirmed **E0432** on a sqlite build. Each page now contradicts itself: line ~897 already says `router` takes `impl Into<rustango::sql::Pool>` and "is not Postgres-only (#1273)".

The same commit updated all four locales for #1502, so the sweep was locale-aware and this one was missed. `Closes #1272` then retires the issue while 3/4 of the affected surface still reproduces it. (Found independently by two angles; upgraded from medium on that plus the compile proof plus the auto-close.)

*Separately:* #1272's first half — scaffolder no longer emitting `admin_router` — is untouched by this PR either way, so `Closes #1272` looks wrong regardless.

**5. `.github/workflows/ci.yml:927` — two more `sqlite_live` targets still print `ok. 0 passed`**

Pre-existing, but this PR fixed one instance of the class and left two identical ones in the job it was already reading. The job builds with `sqlite,tenancy,admin,jobs-postgres,…` and **`jobs-postgres = ["jobs"]` does not enable `postgres`**. It names `--test foreign_key_sqlite_live` and `--test orm_advanced_sqlite_live`, both still gated `all(sqlite, postgres)`.

`foreign_key_sqlite_live.rs` is the clearest: line 1 gates on `postgres`, line 11 carries a **dead** `#![cfg(feature = "sqlite")]`, and the header between them says it proves FK lazy-load works "on any backend *without* Postgres".

Root cause is `foreign_key.rs:391` gating the backend-agnostic `get_pool` on `postgres`, with a v0.35 note saying the gate should have dropped. Neither `every_test_suite_compiles_in_ci` nor `every_live_suite_runs_in_ci` has an invariant for "a `--test` target a job names is cfg'd out under that job's features" — that guard would catch all three at once.

---

<details>
<summary><b>Medium (13)</b></summary>

- **`tests/docs_live_suite_counts.rs:303`** — the new guard survives a gained leading digit. 188→192 **dies** correctly (your verification reproduces), but 188→**1188** passes: the `contains` is unanchored, the exact defect the sibling `claims` array anchors against 25 lines above.
- **`tests/docs_live_suite_counts.rs:304`** — rewrapping a *correct* header across two `//!` lines fails the guard as "stale", sending the reader to recount 188 files.
- **`src/sql/sqlite.rs:98` + `src/migrate/ddl.rs:189`** — the FK-naming correction is right about the emitter it measured and wrong about the one deployments run. `inline_fk_clauses` does emit `CONSTRAINT "{table}_{column}_fkey"` (rendered and confirmed), but system migrations and `testkit::migrate_framework` go through the `SchemaSnapshot` emitter (`diff.rs:1429/1445`), which writes SQLite FKs **anonymously**. So "the framework names every one" is backwards on the `manage migrate` path — and `ddl.rs:189` still asserts the pre-correction claim, so the tree now says both.
- **`src/cache/mod.rs:1031`** — `decode()` promises `None` on a truncated write and returns the prefix. `encode` writes no length for the value, so `decode` returns `body[key_len..]` verbatim; truncating to len-1/-16/-20 returned `Some(prefix)` every time. The PR corrects the outer prose and leaves the inner doc that regenerates it.
- **`CHANGELOG.md:1020`** — the CSRF remedy drops the empty-token guard the framework itself has. `current_csrf_token().unwrap_or_default()` + unconditional `csrf_input_html(&token)` emits `value=""` where `chrome_context_with_session` renders nothing (`admin/helpers.rs:91`). Following it in that state 403s — the exact failure the entry promises to fix. Mirror the `is_empty()` guard.
- **`tests/tx_methods_sqlite_live.rs:9`** — the gate change is **correct** (3/3 pass under `sqlite` alone, no `tenancy` needed), but the comment says the proof never ran "in a sqlite-only **or mysql-only** build". `cfg(feature = "sqlite")` still excludes it from mysql-only, and always will — the file hardcodes `SqlitePool`. Drop the clause. *(Found by three angles.)*
- **`src/migrate/diff.rs:980`** — the recount is also wrong: "two arms further down" → "three arms sit between them", but it is **four** (`RenameTable` :982 → `DropCheckConstraint` :1128). *(Found independently by three angles; upgraded from low on that agreement — a recount PR getting a recount wrong is the defect it exists to prevent.)*
- **`docs/{de,es,fr}/logging.md:170`** — still call `pretty` the default. The `| pretty | … | Default |` row sits four lines from the line changed to `full`, and line ~221 of the same file already says the default is `full`. Each translated page now contradicts itself. The English page has a different table shape and is unaffected.
- **`bin/bump-version.sh:118`** — the corrected shape count at :230 is right, but :161-166 still asserts the verification checks the same shapes the rewriter handles — precisely the belief #1605 exists to record as false.
- **`CHANGELOG.md:857`** — "the key-floor panic no longer names the key's length" is still false at `tenancy/auth_routes.rs:112` and in `SessionSecretError::TooShort`.
- **`src/crypto.rs:81`** — the correction blames the csrf copy for what its own function does. `crypto::constant_time_compare` — the function the comment is attached to — hand-rolls the identical XOR loop, and `crypto.rs` has no `use subtle` at all. `totp` is the only one of the three that actually uses `subtle`.
- **`src/access_log.rs:767`** — no test drives the `access_log=None` mount with a configured redact list. The gap behind finding #1.
- **`CHANGELOG.md:846`** — shipped entries get three different edit treatments in one diff: errata parenthetical ×6, silent rewrite ×3, plus one wholly **new** entry inserted into released 0.57.6. #1520 gets both treatments for the same fact in the same section (:1146 silent, :1479 errata). Worth picking one.

</details>

<details>
<summary><b>Low (14)</b></summary>

Sweeps that stop one site short, and comments that now over-narrate:

- the "two arms further down" phrase survives in `tests/migrate_ddl_tri.rs:316`
- two checkable counts traded for unfalsifiable "some of which" (`dialect.rs:336`, `CHANGELOG.md:893`)
- the commit **body** states #1522 inverted relative to what shipped — the diff and the issue title both say six-not-five
- public `Setup` rustdoc now carries a paragraph about a past rustdoc bug
- the rewritten `constant_time_compare` doc still ships an example that cannot compile
- "the one accessor that is always tenant-scoped" wrongly demotes `Tenant::conn`, which the paragraph two lines above already describes as scoped
- "the arms are authoritative" is policy, not enforcement
- the `macro_internals` guard now documents a blind spot its own failure message denies
- the corrected `without_targets` scope claim is true but unobserved
- the new MySQL 8.0.18 floor is published and never exercised — `docker-compose.yml` pins rolling `mysql:8.0`, so CI never touches the boundary

</details>

---

### Checked and found correct — please don't re-litigate these

- **#1517 error codes: right on all four claims.** Run against live MySQL 8.0.46 — absent `DROP CHECK` → **3821** `ER_CHECK_CONSTRAINT_NOT_FOUND`, absent `DROP FOREIGN KEY` → **1091** `ER_CANT_DROP_FIELD_OR_KEY` (including the composite case the entry singles out) — and PG 16 `DROP CONSTRAINT IF EXISTS` is idempotent for both forms.
- **#1506 landed.** Header table and `by_dialect!` arms now agree; MySQL emits the 8.0.18+ `EXPLAIN ANALYZE` form and 8.0.46 returns `actual time=`.
- **The CSRF findings are not a bypass.** Checked specifically, because the empty-token recipe and `constant_time_eq(b"", b"")` returning `true` look like they compose. They don't: `verify_form_token` requires `(Some(cookie), Some(form))` (`csrf.rs:693`), and when `current_csrf_token()` is `None` the layer set no cookie — so the pair is `(None, Some(""))` → reject. The outcome is a 403, not a silent pass.
- **Comment density is not a new pattern.** The "This used to say…" errata shape is pre-existing house style (7 instances at `188eeada`); this PR adds 9 more at comparable length. Only the one on `Setup`'s *public* rustdoc is flagged, and only because it is user-facing.
- **#1516 correctly *not* closed** — it's open and is a live code defect; the PR documents without claiming a fix. Same correct treatment for #1530, #1535, #1605.
- **`sqlite,admin` is a real `feature_combos` row** (`ci.yml:699`) and `admin` pulls neither `manage` nor `tenancy`, so it genuinely exercises the gate. `postgres,admin` is indeed not in the matrix.
- **No `search_path` leak in `Tenant::pool()`** — the schema-mode pool is scoped via `PgConnectOptions::options` at connect time, not a session-level `SET`, so the #1224 hazard doesn't apply. Cache keys are `org.slug`, which is `unique` — no cross-tenant collision possible.
- **The `#1461` citation in `migrate/diff.rs` is defensible** — it refers to the tri-dialect conversion epic. Nearly filed as a bad citation, withdrawn.
- **`Format::Full` is the default**; `.with_target` does precede the format match in both install paths; `Pool::connect` exists and the English `viewsets.md` snippet compiles verbatim in a downstream crate.

### Outside this change

Pre-existing, not caused here, worth their own issues:

- **`forms/csrf.rs:674`** — the admin's effective CSRF cookie is set without `Secure` and ignores `CsrfConfig`. The most security-relevant thing found that this PR didn't cause.
- **`forms/csrf.rs:701`** — `CsrfLayer` treats an empty cookie and an empty token as a match (pinned by a test). Hardening gap, not currently reachable, see above.
- **`ci.yml:264`** — the `guards` job comment says fifteen; the job lists eighteen.
- **`diff.rs:1445`** — the inline composite-FK branch in the snapshot emitter looks unreachable (filed `likely`, not confirmed).
- **Registry-scoped models resolve through `t.pool()`** — the baked path is `<schema>,public`, so a `scope="registry"` model returns every tenant's rows. Pre-existing, deliberate, identically true of `Tenant::conn` — flagged only because the new prose leans harder on "always tenant-scoped".

### Coverage and limits

**Backends actually exercised:** MySQL 8.0.46 and PostgreSQL 16 (real containers, real DDL probes), SQLite 3.51.0 (CLI grammar probes plus two renders through `create_table_sql_with_dialect` and `render_changes_split_with_dialect`). Built and ran `--features sqlite` and `sqlite,tenancy`; replicated the `sqlite_live` job command across its 60 targets.

**Mutation-tested:** the new 188-suite guard was driven with four mutations (two it failed to catch or caught wrongly); the `tx_methods` gate was reverted and watched to die as `ok. 0 passed`.

**Gaps, stated plainly:** the MySQL **8.0.18 floor** is read-only — no 8.0.17 server exists to test against, and CI pins the rolling `mysql:8.0` tag, so nothing touches the boundary. The tenancy findings are inspection-only (the corroborating suite needs a live Postgres with seeded tenants). Nobody opened the four translated `testing.md` pages, `docker/soak/`, or examples beyond `cookbook_blog`'s call sites. Earlier-release CHANGELOG assertions rewritten here were spot-checked, not exhaustively re-derived.

Issue #1573, opened 2026-09-17
media: two more write-path defects, and an on_delete suppression that blocks #1557


ujeenet:
From the crew review of the 0.57.7 release (run `pr-1544`): correctness-002, -003, -004, tenancy-003, security-003, dialects-003.

The release's blocking write-path regressions are fixed on the release branch. These are the ones that are real but not regressions, or that need more than a patch release should carry.

## 1. `purge` is the one write path in the batch left non-transactional — `mod.rs:635`

It gained the same unlink-then-delete pair as `purge_pending`, but as two separate `raw_execute_pool` calls. Its **new** rustdoc promises "the row is left in place in every error case" — true of the row, false of the tag links and of the storage object it has already deleted. A failure between the two leaves reclaimed links for a row that still exists.

## 2. `delete_collection` resolves its subtree outside its own transaction — `mod.rs:1007`

`collect_descendant_ids` runs before `transaction_pool`, so a collection created under the subtree mid-delete survives with a `parent_id` pointing at a soft-deleted parent — the exact dangling state the release added the subtree walk to remove.

## 3. `on_delete` `None → Some` is suppressed unconditionally — `diff.rs:615`

```rust
if p.on_delete.is_some() && p.on_delete != c.on_delete { ... }
```

`None → Some("CASCADE")` is silent **forever**, while the mirror `Some → None` hard-errors. The suppression was deliberate — it is what stops an upgrade with zero model changes failing `make_migrations` — but it was written as a one-time upgrade allowance and behaves as a permanent one.

**This blocks #1557.** Adding an `AlterFkOnDelete` operation would never fire for the upgrade case, because detection is suppressed in exactly that direction. #1557 needs this resolved first, not alongside.

## 4. Nothing ties `MediaPerms`'s pool to the manager it authorizes — `router.rs:500`

`MediaPerms::new(pool)` and `MediaManager::new_pool(pool, …)` take separate pools with nothing checking they agree. Under tenancy the wrong one means permissions read from a different tenant than the rows. `pool_dyn()` is public, so a `MediaPerms::from_manager(&mgr)` constructor would make the relationship structural rather than a convention in prose.

## 5. `rustango_media.change` creates tag rows — `router.rs:442`

[...]

Issue #1557, opened 2026-09-17
on_delete reaches new databases only: migrate leaves an existing one uncorrected and exits 0


ujeenet:
#1549 made `on_delete` reach the database. It reaches **new** databases only — an existing one keeps the constraints it already has, and nothing tells the operator.

Follow-up to #1549 / PR #1555, deliberately deferred out of that PR so a patch release did not grow a schema-rewrite operation.

## What happens today

Reproduced end to end by scaffolding an app on `release/v0.57.7` at `6d4a7fe6`, migrating it, then repinning to the fix and rebuilding:

```
$ cargo build                     # clean, no warnings
$ cargo run -- migrate
nothing to migrate (already up to date)          rc=0
$ ls system/migrations/                           # unchanged
```

And the database it leaves behind:

```sql
CREATE TABLE "item_note" (... "item_id" INTEGER NOT NULL REFERENCES "item" ("id"))
```

No `ON DELETE CASCADE`, on the user's own FK *or* the framework's. A greenfield project on the same commit gets it correctly, so this is specifically upgrade-vs-new.

## Why

`detect_changes` turns a changed `on_delete` into no schema operation, so nothing is emitted for a table that already exists. That is correct for what the release could safely do — but it means the fix's headline is true only for new databases.

The failure mode is the bad kind: **silent and exit 0.** `UPGRADING` step 7 asks the reader to look for a generated system migration, they find none, and "none" reads identically as "you are fine" and "the framework did not notice".

## Scope

Eleven framework FKs declare `cascade` — ten of them in `tenancy` (roles, permissions, agent skills), one in media — and `fold_in_framework_tables` puts them in every project's snapshot. So this reaches **every existing tenancy project**, not only apps that use media.

## What exists now

`UPGRADING.md`'s 0.57.7 section states the limitation, gives the `ALTER`, and gives the catalog queries to check what you actually have:

- PostgreSQL — `SELECT conname, confdeltype FROM pg_constraint WHERE contype='f'` (`a` = NO ACTION, `c` = CASCADE, `n` = SET NULL)
- MySQL — `SELECT constraint_name, delete_rule FROM information_schema.referential_constraints`

That is a workaround, not a fix.

## The work

Emit a drop/add foreign-key pair when a snapshot's `on_delete` changes, so `migrate` corrects an existing database the way it corrects anything else.

Known constraints:

[...]

Issue #1549, opened 2026-09-17
on_delete is silently dropped by every system migration: declared cascade becomes NO ACTION in the database


ujeenet:
Every `on_delete = "…"` declared on a model is silently discarded by the migration path that production actually runs. The declared referential action never reaches the database, and a declared `cascade` becomes a hard **refusal** at runtime.

Found during a review of #1545. Not media-specific — media is just where it was noticed.

## Proven, against live databases

`MediaTagLink.tag_id` has declared `on_delete = "cascade"` since `8daa2a66` (2026-08-06):

```rust
#[rustango(fk = "rustango_media_tags", on = "id", on_delete = "cascade")]
```

What is actually in a database built by rustango's own system migrations:

```
MySQL:  CONSTRAINT `rustango_media_tag_links_tag_id_fkey`
        FOREIGN KEY (`tag_id`) REFERENCES `rustango_media_tags` (`id`)   ← no ON DELETE
PG:     conname = rustango_media_tag_links_tag_id_fkey, confdeltype = 'a'  ← 'a' = NO ACTION
```

**All 7** framework FK constraints in the live PostgreSQL carry `confdeltype='a'`. The consequence, executed:

```sql
DELETE FROM rustango_media_tags WHERE id = 1;
ERROR 1451 (23000): Cannot delete or update a parent row: a foreign key constraint fails
```

So the declaration says "deleting a tag removes its links" and the database says "you may not delete a tag."

## The mechanism

1. The macro parses and validates `on_delete` (`rustango-macros/src/lib.rs:10633-10651`) and threads it into `FieldSchema::fk_on_delete`. ✅
2. `migrate/ddl.rs:260` and `:327` emit `ON DELETE …` correctly — **from a `ModelSchema`**. ✅
3. `RelationSnapshot` (`migrate/snapshot.rs:195-200`) is exactly:
   ```rust
   pub struct RelationSnapshot {
       pub kind: String,   // "fk" or "o2o"
       pub to: String,
       pub on: String,
   }
   ```
   **There is no field for `on_delete`.** The value is dropped the moment a snapshot is built. ❌
4. System migrations and `testkit::migrate_framework` both go `SchemaSnapshot::from_models` → `detect_changes` → `render_changes_split_with_dialect`. That is path 3, so the clause is never emitted.

[...]

Issue #1533, opened 2026-09-17
ContentType cache is process-global and keyed without the tenant — documented as a feature


ujeenet:
**Severity:** Medium · **Area:** `contenttypes` · **Found on:** `develop` @ `ca726393`

The `ContentType` cache is process-global and keyed on `(app_label, model_name)` only. The pool the row came from is not part of the key, and the doc states this as an intended property.

[contenttypes.rs:226-242](crates/rustango/src/contenttypes.rs#L226-L242):

> The cache is keyed by the natural key, not by Rust `TypeId`, so cached entries survive **across tenants** / rebinds / hot-reloads

```rust
fn cache() -> &'static std::sync::RwLock<HashMap<(String, String), ContentType>> { ... }
```

and [:265-291](crates/rustango/src/contenttypes.rs#L265-L291):

```rust
pub async fn get_by_natural_key(pool: &Pool, app_label: &str, model_name: &str)
    -> Result<Option<Self>, ExecError>
{
    let key = (app_label.to_owned(), model_name.to_owned());
    if let Some(hit) = cache().read()...get(&key).cloned() { return Ok(Some(hit)); }
```

`pool` is ignored on the hit path.

## Why that is wrong under database-mode tenancy

In database-mode each tenant has its own database, therefore its own `rustango_content_types` table, therefore its own auto-increment ids. The same `("blog", "post")` natural key is id 7 in one tenant and id 12 in another, depending only on seeding order.

So the first tenant to warm the cache decides the id every other tenant gets. The cached struct carries that `id`, and content-type ids are the discriminator for:

- generic foreign keys (`content_type_id` + `object_id`) — a GFK written under the wrong id points at the wrong model
- the permissions tables, which key on `content_type_id`

A GFK row written for tenant B with tenant A's id is silently mis-typed, and a permission row matched on the wrong content type is an authorization decision made against the wrong model.

## Scope, honestly

`get_for_model` / `get_by_natural_key` have **no callers inside the framework** — `grep -rn 'get_for_model\|get_by_natural_key' --include='*.rs' crates/rustango/src` outside `contenttypes.rs` is empty. So nothing rustango ships is currently broken by this. It is a public, documented API in a framework whose headline feature is multi-tenancy, and the doc actively advertises the cross-tenant sharing as a feature.

## Suggested fix

Either:

[...]

PR #1616, opened 2026-09-19
fix(sqlite): auto_now_add timestamps that compare and sort (#1464)

### LINKED ISSUE #1464: SQLite: auto_now_add columns are stored in a format that Rust-bound DateTime cannot compare against
## Summary

On SQLite, an `auto_now_add` column is written by SQLite's own
`CURRENT_TIMESTAMP` but compared against a value bound from Rust as
RFC3339. The two text formats diverge at position 10, so **every
comparison against such a column is wrong**.

```
stored by DEFAULT CURRENT_TIMESTAMP : 2026-09-15 02:53:25
bound   by sqlx for DateTime<Utc>   : 2026-09-15T02:53:25.123456+00:00
                                                ^
                          ' ' (0x20) < 'T' (0x54)
```

`WHERE placed_at < $1` is therefore **true for every row**, whatever the
bound value.

## How it was found

The commerce soak (`docker/soak/`), running cursor pagination on
`Order.placed_at`. Postgres and MySQL page correctly; both SQLite
instances return page one forever:

```
ok    #1459  timestamp cursor advances [single-pg]   5 then 25 distinct rows
ok    #1459  timestamp cursor advances [single-my]   5 then 25 distinct rows
FAIL  #1459  timestamp cursor advances [single-sq]   page two repeats rows from
                                                     page one: [18322, 18323,
                                                     18324, 18325, 18326]
```

Confirmed directly against the container's database:

```sql
sqlite> .schema commerce_order
CREATE TABLE "commerce_order" (..., "placed_at" TEXT DEFAULT CURRENT_TIMESTAMP NOT NULL, ...);

sqlite> SELECT id, placed_at FROM commerce_order ORDER BY placed_at DESC LIMIT 3;
18334|2026-09-15 02:53:25
18333|2026-09-15 02:51:39
18332|2026-09-15 02:51:17
```

## This is already known, and already worked around once

`audit.rs` carries the diagnosis verbatim and patches it at that single
call site:

```rust
// #560 — `occurred_at` is `TEXT DEFAULT CURRENT_TIMESTAMP` on
// SQLite (`CREATE_TABLE_SQL_SQLITE`). SQLite's CURRENT_TIMESTAMP
// emits `"YYYY-MM-DD HH:MM:SS"` (space sep, no fractional, no
// timezone); sqlx-sqlite would otherwise encode a
// `chrono::DateTime<Utc>` as RFC3339, and lex-compare diverges
// at position 10 (space < T). Bind the SQLite cutoff in the
// same CURRENT_TIMESTAMP shape.
let bind = if pool.dialect().name() == "sqlite" {
    SqlValue::String(cutoff_ts.format("%Y-%m-%d %H:%M:%S").to_string())
} else {
    SqlValue::DateTime(cutoff_ts)
};
```

That fixes audit cleanup and nothing else. Cursor pagination is the
second victim; every other `WHERE <auto_now_add column> <op> $1` on
SQLite has the same defect, silently.

## Scope

Not only comparison. A SQLite database today holds a **mixture** of
formats — rows whose timestamp came from the `DEFAULT` are space-
separated, rows written with an explicit Rust value are RFC3339 — so
`ORDER BY` across the two is wrong as well.

## Options

1. **Central bind fix.** Make `bind_match_sqlite!` encode
   `SqlValue::DateTime` as `%Y-%m-%d %H:%M:%S`. One place, fixes every
   comparison, lets the `audit.rs` special case go. Loses sub-second
   precision on SQLite, and rows already written as RFC3339 then
   mis-compare the other way.
2. **DDL fix.** Stop emitting `DEFAULT CURRENT_TIMESTAMP` for
   `auto_now_add` on SQLite — emit `strftime('%Y-%m-%dT%H:%M:%fZ','now')`,
   or bind the value from Rust on insert. Keeps sub-second precision;
   new tables become consistent, existing databases do not.
3. **Either, plus a migration** that normalises existing SQLite datetime
   columns. Most correct, and the only option that fixes databases that
   already hold both formats.

Whichever is chosen, the guard has to be a comparison that actually
pages — a fixture whose timestamps are distinct and Rust-written will
pass while the bug is fully present, which is why unit tests have not
caught this.

## Not blocking 0.57.5

Filed rather than fixed in the 0.57.5 branch: every option changes the
stored datetime format, which wants its own release and a migration
note. The soak reports the two SQLite legs as a known gap, named to this
issue, so they stay visible without being mistaken for a pass.

ujeenet:
First of 0.57.11. Closes #1464 — the only `critical` in the backlog.

## The defect

SQLite has no datetime type, so a `DateTime<Utc>` column is TEXT and compares
**lexicographically**. `auto_now_add` requires `Auto<..>`, so the macro skips the column on
INSERT and the DB default fires — and that default was `CURRENT_TIMESTAMP`, which writes
`YYYY-MM-DD HH:MM:SS` while sqlx binds RFC3339. They diverge at position 10, `' '` (0x20)
against `'T'` (0x54), so `WHERE col < ?` was true for **every** row whatever was bound.

Cursor pagination returned page one forever. The commerce soak found it; no unit test could,
and the issue predicted exactly why.

## The decision the issue asked for first

Fix the DDL **and** convert what is already stored. Either alone leaves a database holding
both shapes, which sorts wrongly with no bind involved at all.

The format was chosen by measuring sqlx rather than assuming. Its encoding is **variable
width** — chrono's `%.f` emits 0, 3, 6 or 9 fractional digits, so a zero fraction has no
fractional part at all:

```
2027-01-15T08:00:00+00:00           (no fraction, 25 chars)
2027-01-15T08:00:00.123+00:00       (3)
2027-01-15T08:00:00.413181+00:00    (6)
```

That family is still sort-sound, because `'+'` (0x2B) precedes `'.'` (0x2E) and every digit:
no-fraction precedes any fraction, and a short fraction precedes a longer one extending it.
The default just has to join the family, so it is
`(strftime('%Y-%m-%dT%H:%M:%f000+00:00','now'))`. Parens are required — SQLite takes a
non-constant DEFAULT only in expression form.

## Shape

| file | change |
|---|---|
| `sql/sqlite.rs` | `translate_default_expr` emits the new shape; two shared constants own the format and the legacy mask |
| `migrate/sqlite_datetime.rs` | new — converts stored rows, driven by the `ModelEntry` registry so **user** models are covered, plus an explicit list for the framework tables that are not model-derived |
| `migrate/manage.rs` | runs on every `migrate`, not behind a repair flag |
| `runner.rs`, `i18n/db.rs` | the hand-written SQLite DDL |
| `audit.rs` | #560's bind-side workaround deleted; both sides agree now |

**Not behind a repair flag** for the reason the bug demonstrates: an operator does not know
to run a verb they have not heard of, and silence reads exactly like "you are fine".
Idempotent — the mask stops matching once a value is converted.

`Date` and `Time` are excluded from the sweep deliberately: they carry no separator to
disagree about, and sweeping one would turn a correct date into a timestamp.

## One regression this introduces, and closes

Removing the audit workaround opens a window between upgrading the crate and running
migrations where a legacy row is "less than" everything. Elsewhere that window costs a wrong
query result — the old bug continuing. In `audit`'s retention sweep it is a **DELETE**, so it
would destroy history. That comparison therefore normalises the *stored* side in a `CASE`
rather than relying on the migration having run. It costs the index on a path that scans by
definition.

## Verification

Every guard revert-drilled — fix undone, guard watched to **fail**, then restored.

- `auto_now_add_sqlite_format` builds its table through `testkit::create_tables_for` and
  inserts **without** the timestamp, so the value under test is the one the database wrote.
  Seeding it explicitly passes against unfixed code, which is why the existing
  `cursor_pagination_on_a_timestamp` never caught this.
- `sqlite_datetime_normalise` seeds the legacy shape deliberately. **Five of its seven failed
  the first drill and two did not** — one was testing a proxy: its cutoffs were a day either
  side of the rows, so the comparison decided on the date at position 9 and never reached the
  separator. Same-day cutoffs now force it. That is recorded in the test's own doc comment.
- `sqlite_datetime_ddl_is_consistent` reads the source, because the next hand-written table
  is a plain string with nothing to stop it. Keyed on TEXT + a bare `CURRENT_TIMESTAMP`: the
  first version matched the whole line and flagged the translations table's **MySQL** DDL,
  which is correct as written. Registered in the `guards` job — `every_structural_guard_runs_in_ci`
  failed until it was.

573 test targets pass under `sqlite,tenancy`, 0 failures. `postgres`, `sqlite`, `mysql` and
`sqlite,tenancy` all build warning-free.

**Not run locally:** `cargo test --workspace --all-features --no-run` — it exhausted the disk
twice. `tests_compile` (ungated, ci.yml:379) runs exactly that on this PR.

## Sequencing note for the rest of 0.57.11

**#1573 item 3 has moved out of this release**, and the roadmap was wrong to put it here.
Un-suppressing `None → Some` without an `AlterFkOnDelete` operation makes `make_migrations`
hard-error for every existing project. Demonstrated, not argued — removing the guard fails
`an_upgrade_is_not_a_schema_change` with:

```
`fkod_post_cascade.author_id` fk on_delete changed: None → Some("CASCADE")
`fkod_post_set_null.author_id` fk on_delete changed: None → Some("SET NULL")
```

Every snapshot on disk predates #1549, and `fold_in_framework_tables` puts eleven framework
FKs into every project's snapshot. The advice that refusal prints cannot be followed, because
the operation does not exist yet. It belongs in the #1557 PR, as its own commit ahead of the
emitter. 0.57.11 is this plus #1533.


ujeenet:

PR #1618, opened 2026-09-20
B: stop the sweep rewriting rows the fix already wrote

ujeenet:
**Blocker 2 of 7** from the `pr-1616-rework` review — stacked on #1617.

## The defect

The sweep asked `col <> strftime(FMT, col)` to mean *"not already canonical"*. It does not.

SQLite's `%f` is **milliseconds**. The bind path's chrono `%.6f` is **microseconds**. So
`strftime` is not the identity on a value the fix itself writes:

```
stored by the bind path : 2026-09-20T08:00:00.413681+00:00
strftime renders it as  : 2026-09-20T08:00:00.414000+00:00
```

Roughly **999 in 1000 correct rows** matched the predicate. Every `migrate` rewrote them,
rounded each up to 500 µs forward, and reported them as "written by the pre-0.57.11
default". `dialects` measured nine rows written 200 µs apart collapsing to three distinct
values.

## Why I missed it

The module doc claimed "idempotent by construction" — asserted, never checked. And the
premise probe I ran *before* writing the shim tested bind → store → decode → bind. It
never tested store → **sweep** → compare. I verified the property I had thought of, which
is the same failure this whole issue keeps producing.

## The fix

`SQLITE_CANONICAL_GLOB` asks the question directly rather than round-tripping. `?` is one
character and `[0-9]` a digit, so it matches the canonical spelling and nothing else.

Verified across every case that matters:

| input | result |
|---|---|
| canonical, microseconds | **preserved** |
| canonical, whole ms | preserved |
| legacy space-separated | converted |
| old variable-width RFC3339 | converted |
| space + fraction | converted |
| unparseable junk | untouched |
| second pass | **0 rows** |

## The test uses microseconds on purpose

A whole-millisecond fixture passes against the *broken* predicate too — it would have
been another guard that cannot fail, which is exactly what #1617 exists to clean up.

Drilled: against the old predicate it fails with *"the sweep rewrote a row the bind path
had already written correctly"*.

## Also

The module doc still described the `LIKE` mask deleted two commits earlier.

575 targets pass; `postgres`, `sqlite`, `mysql`, `sqlite,tenancy` all warning-free.

## Rest of the stack

C — audit over-deletes · D — four missed writers · E — `no such column` aborts migrate ·
F — old tables keep the old DEFAULT · G — `keep_last_n`


ujeenet:

PR #1619, opened 2026-09-20
C: normalise instead of enumerating in the retention DELETE

ujeenet:
**Blocker 3 of 7** from the `pr-1616-rework` review — stacked on #1618.

## The defect

The second leg matched `NOT LIKE '____-__-__ __:__:__'` — an *exactly 19-character*
space-separated value. So this spelling walked straight past it:

```
2026-09-20 08:00:00.123456
```

That is the PG/MySQL dump spelling, and what sqlx writes for a bound `NaiveDateTime`.
With the guard skipped, the bare range judged it — and `' '` (0x20) sorts below `'T'`
(0x54), so every same-date legacy row is "older" than a canonical cutoff whatever its
clock time. **Rows hours after the cutoff were deleted.**

Enumerating the shapes I could think of, in the guard written to stop exactly that.

## The fix

```sql
-- was
AND (occurred_at NOT LIKE '____-__-__ __:__:__' OR occurred_at < ?)
-- now
AND strftime(FMT, occurred_at) < ?
```

Normalises the stored side rather than recognising spellings: correct in every shape,
fails closed on unparseable text. The leading indexed range is untouched, so the plan
stays `SEARCH … USING INDEX (occurred_at<?)` and the measured win over the old `CASE`
(2/3/5 ms against 186/2969/1961 ms) stands.

`SQLITE_LEGACY_DATETIME_LIKE` had its last reader removed here — #1618 moved the sweep to
a `GLOB` shape test — so it is deleted rather than left dead.

## The test caught me twice, and that is worth reading

Its **first** fixture seeded rows dated *tomorrow*. The comparison then decides on the
date at position 9 and never reaches the separator at position 10, so it passed against
the broken predicate.

That is the identical mistake fixed in `a_converted_row_compares_against_a_bound_timestamp`
earlier today — repeated while writing the test *for* that class of bug.

Now same-day, later-hour. And the fraction is deliberate: a bare 19-character value
survives the old predicate too, so the obvious fixture would have proved nothing.

Drilled: fails against the old predicate, passes against the fix.

575 targets pass; `postgres`, `sqlite`, `mysql`, `sqlite,tenancy` warning-free.

## Rest of the stack

D — four missed writers · E — `no such column` aborts migrate · F — old tables keep the
old DEFAULT (needs a table rebuild) · G — `keep_last_n`, gated on F


ujeenet:

PR #1621, opened 2026-09-20
D: route the last four writers through the shared binders

ujeenet:
**Blocker 4 of 7** from the `pr-1616-rework` review — stacked on #1619.

#1616 claimed four writers and one canonical shape. Reviewers found **eight**, and the
count rose every round, because nothing structural stopped the next one. This closes the
four that were missed — net **−93 lines**.

## `audit` held 116 lines of hand-copied macro expansion

`bind_value_pg` / `_my` / `_sqlite` are arm-for-arm copies of `bind_match_pg!` /
`_mysql!` / `_sqlite!`. When #1464 moved SQLite datetime encoding into the macro, every
call site picked it up **except the copy** — which is the whole hazard: a copied
expansion is invisible to every test that exercises the macro it drifted from. Two
reviewers found it independently.

The macro is module-scoped to `sql::executor`, so the fix is the wrapper functions that
already existed one layer up — `bind_query`, `bind_query_my`, `bind_query_sqlite` —
widened `pub(super)` → `pub(crate)` and re-exported. All three copies deleted.

## `jobs/pg.rs` bound `now.to_rfc3339()`

Variable-width, and used **twice**: written into `locked_at`, and compared against
`run_at`, whose DDL default this release changed to the fixed six-digit shape.

A whole-second `now` encodes with no fractional part at all, and `+` (0x2B) sorts below
`.` (0x2E) — so `run_at <= ?` read a **due job as not-yet-due**.

## `admin/views.rs` `BucketBind::DateTime` bound raw

The date hierarchy compared a variable-width bound against a fixed-width column.
`BucketBind::Date` is deliberately untouched — a bare date has no separator to disagree
about.

## Checked and left alone

`core/value.rs`'s `to_display_string`, whose own doc says *"NOT for SQL emission"*; and
`humanize` / `dateformat` / admin JSON, which are presentation.

## What this does and does not settle

The **value** path is now structural: one encoder, one set of binders, reached by
everything. That is the shape the review asked for.

What remains is SQL **text**, where `Dialect::translate_default_expr` already renders the
canonical default byte-for-byte and three hand-written DDLs duplicate it instead. That is
a separate change and belongs with F.

## Verification

4783 tests under nextest + 68 doctests, all passing. `postgres`, `sqlite`, `mysql`,
`sqlite,tenancy`, `postgres,jobs`, `sqlite,admin` all warning-free.

## Rest of the stack

E — `no such column` aborts migrate · F — old tables keep the old DEFAULT (table rebuild) ·
G — `keep_last_n`, gated on F


ujeenet:

PR #1631, opened 2026-09-21
Trim the GHSA-c4gg prose to what a caller needs

ujeenet:
Stacked on #1630. No behaviour change.

Three doc blocks on the auth middleware ran 31, 21 and 20 lines and
mostly re-told the vulnerability. That story belongs in the commit
message, the CHANGELOG and the advisory — all three of which have it.
An API doc's job is to tell a caller what to do.

| Block | Before | After |
|---|---|---|
| `RouterAuthExt` | 31 | 14 |
| `tenant_pool` | 21 | 9 |
| module example note | 3 | 0 |
| `auth_binds_to_the_request_tenant` header | 18 | 6 |

`RouterAuthExt` now leads with the migration — drop the argument —
instead of three paragraphs of history, and keeps one sentence on why
no variant takes a pool. `tenant_pool` keeps the two things a reader
of *that function* needs: why it is backend-erased, and that it fails
closed.

134 added comment lines against `develop` down to ~100, on 462 added
lines. The five auth tests pass unchanged.

Two sibling PRs carry the same weight and are not touched here,
because their files are not on this branch: #1629 is 42% comments
(199 of 471, including a 27-line block on the ContentType cache key)
and #1628 is 34%. Worth the same pass before they merge.


ujeenet:
Superseded by the cherry-pick onto the current `develop`. #1630 was rebase-merged, so this branch's base commits no longer exist and the PR could not merge cleanly. Re-applied rather than rebasing a pushed branch.

